PR #20, opened 2018-04-17
Jose Garza

jggarza5:
Javascript II

### DIFF
--- a/assignments/array-methods.js
+++ b/assignments/array-methods.js
@@ -89,7 +89,21 @@ console.log(ticketPriceTotal);
 // Now that you have used .forEach(), .map(), .filter(), and .reduce().  I want you to think of potential problems you could solve given the data set and the 5k fun run theme.  Try to solve 3 unique problems using one or many of the array methods listed above.
 
 // Problem 1
+let emailList =[];
+runners.forEach(function (runner) {
+    emailList.push(runner.email)
+});
+console.log(emailList);
 
 // Problem 2
-
+let fullNamePlusCompany = [];
+runners.forEach(function (runner) {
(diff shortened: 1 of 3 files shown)


assignments/array-methods.js
```diff
@@ -56,28 +56,54 @@ const runners = [{"id":1,"first_name":"Charmain","last_name":"Seiler","email":"c
...
 let fullName = [];
+runners.forEach(function (runner) {
+    fullName.push(runner.first_name + " " + runner.last_name);
+});
 console.log(fullName);
 
 // ==== Challenge 2: Use .map() ====
 // The event director needs to have all the runner's first names converted to uppercase because the director BECAME DRUNK WITH POWER. Convert each first name into all caps and log the result
-let allCaps = [];
+let allCaps = runners.map(function(elem, index, array) {
```
enitchals, line 66: Using `function(elem, index, array)` (rather than more descriptive names) is likely helpful if you intend to come back to this as example code, but it's a good habit to use your own variable names that are more descriptive.

Also, remember that the only required value for map/reduce/filter/forEach is the `elem`, representing the current item being iterated over -- so if you don't plan to use `index` or `array`, there's no need to include them (unless it's just for practice). It will still work fine, but it might be confusing or make your code appear incomplete if you pass values without using them.

assignments/closure.js
```diff
@@ -1,15 +1,37 @@
 // ==== Challenge 1: Write your own closure ====
 // Write a simple closure of your own creation.  Keep it simple!
 
+const challenge1 = (first_name, last_name) => {
```
enitchals, line 4: You did a great job making some progress on this Closure material. I think you'll get it down with practice :+1:

enitchals (commented):
Great work! I know you had some trouble with Closure, but don't worry -- it will click.

PR #24, opened 2018-04-17
Xang Thao javascript II

P-web-art:


assignments/array-methods.js
```diff
@@ -54,30 +54,57 @@ const runners = [{"id":1,"first_name":"Charmain","last_name":"Seiler","email":"c
...
 console.log(largeShirts);
 
 // ==== Challenge 4: Use .reduce() ====
 // The donations need to be tallied up and reported for tax purposes. Add up all the donations into a ticketPriceTotal array and log the result
-let ticketPriceTotal = [];
+let ticketPriceTotal = runners.reduce(function(a, b){
+    return a += b.donation;
+}, 0);
+console.log("ticket price total");
 console.log(ticketPriceTotal);
```
enitchals, line 86: I can't remember if you've run into this yet, but you can always console.log multiple arguments: `console.log("ticket price total", ticketPriceTotal);`

assignments/array-methods.js
```diff
@@ -54,30 +54,57 @@ const runners = [{"id":1,"first_name":"Charmain","last_name":"Seiler","email":"c
...
+  return runner.company_name;
+});
+
+console.log(runnersCompanies);
 
 // Problem 2
 
-// Problem 3
\ No newline at end of file
+runners.forEach(function(item, index) {
```
enitchals, line 100: Nice work with these creative problems :+1:

assignments/closure.js
```diff
@@ -1,11 +1,37 @@
 // ==== Challenge 1: Write your own closure ====
 // Write a simple closure of your own creation.  Keep it simple!
 
+let userCreater = function(password) {
+  let encryptPass = password.split("").reverse().join(""); // encrpyted password
```
enitchals, line 5: I love that you're already thinking about encryption!

assignments/closure.js
```diff
@@ -16,4 +42,27 @@ const counterFactory = () => {
   // Return an object that has two methods called `increment` and `decrement`.
   // `increment` should increment a counter variable in closure scope and return it.
   // `decrement` should decrement the counter variable and return it.
+  let count = 0;
+  return {
+    increment: function() {
```
enitchals, line 47: Both `increment` and `decrement` should return the variable after changing it. You can do this by returning count after incrementing/decrementing, or you can use `return ++count` as a one-liner, which increments `count` before returning it.

enitchals (commented):
Great work! I've made some suggestions for refactoring, but just for future reference -- no need to go back to this unless you want to.

PR #30, opened 2018-04-17
lonnie davis

ldavis0866:


JacquelynnRohrich:
Hey Lonnie, I checked out your work and I think you did a fantastic job. Thank you for commiting with detail. I also appreciate you naming it with just your first and last name makes my life easier when searching for you guys!  Keep up the amazing work and I challenge you to go back in the future and finish that last one up :)

PR #43, opened 2018-04-17
Usman Javed

ujaved931:
### Function Conversion
Convert the ES5 syntax into ES6 arrow syntax
- [x] myFunction
- [x] anotherFunction
- [x] add
- [x] subtract
- [x] triple

### Array Methods
- [x] Challenge 1: Use .forEach()
- [x] Challenge 2: Use .map()
- [x] Challenge 3: Use .filter()
- [x] Challenge 4: Use .reduce()
- [x] (1/3) Challenge 5: Be Creative
- [x] (2/3) Challenge 5: Be Creative
- [x] (3/3) Challenge 5: Be Creative

### Closures
- [ ] Challenge 1: Write your own closure
- [x] Challenge 2: Create a counter function
- [x] Challenge 3: Create a counter function with an object that can increment or decrement

## Good Habits
- [x] Pull Request title is your first and last name (add any notes in the comment field)
- [x] Code has been committed regularly
- [x] Commit comments are informative
- [x] Code is organized in a logical way

### DIFF
--- a/assignments/array-methods.js
+++ b/assignments/array-methods.js
@@ -33,7 +33,7 @@ const runners = [{"id":1,"first_name":"Charmain","last_name":"Seiler","email":"c
 {"id":29,"first_name":"Morrie","last_name":"Rainard","email":"mrainards@yale.edu","shirt_size":"XS","company_name":"Podcat","donation":52},
 {"id":30,"first_name":"Fidel","last_name":"Roskelly","email":"froskellyt@ibm.com","shirt_size":"XS","company_name":"Avavee","donation":5},
 {"id":31,"first_name":"Toni","last_name":"MacSweeney","email":"tmacsweeneyu@parallels.com","shirt_size":"M","company_name":"Jaloo","donation":82},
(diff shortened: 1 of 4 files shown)


assignments/array-methods.js
```diff
@@ -54,30 +54,85 @@ const runners = [{"id":1,"first_name":"Charmain","last_name":"Seiler","email":"c
...
+//      }
+//      else {
+//        return item.emailFormat === "@"
+//      }
+//    }
+
+// }
+let wrongEmail = runners.filter((item, index, array) => {
+let emailFormat = [];
+  for(var i = 0; i < emailFormat.length; i++) {
```
nphillips78, line 116: ES6 uses 'let' instead of 'var'.

nphillips78, line 116: If I understand what you're trying to do - you're trying to check the list for any emails that aren't formatted correctly? You're on the right track with your if statement. I can't think of a way to check the email formats without using regex - you need to check for more than just the `@` to really verify formats. You might try the `contains` method to check for just that `@` which would get you close enough for now. You can always come back later when you've learned regex.

ujaved931: That's exactly what Tramane was telling me, he had a different method as well, which was quite use full. I'll have a try again over the weekend, maybe look into regex by myself. Thanks for the feedback Cole. :)

PR #68, opened 2018-05-22
Adrian Adames CS11 - JavaScript-II

adrianadames:


sparekh315 (commented):
Your code looks great. For Function Conversion -the last problem - the solution code has it listed as a constant, but your code still passes.

PR #76, opened 2018-05-22
Jennifer Player: Javascript II

chainchompa:


metten0:
So far so good! Might I suggest committing a bit more often? A good general rule is to commit every time a significant portion is completed or a breakthrough is achieved. Keep up the great work!

PR #90, opened 2018-05-22
Katie Gorbell - Javascript-II

kgorbell:


assignments/function-conversion.js
```diff
@@ -1,23 +1,15 @@
 // Take the commented ES5 syntax and convert it to ES6 arrow Syntax
 
-// let myFunction = function () {};
+let myFunction = function () {};
```
holly-g: With no parameters, you can express the arrow function like so: `let myFunction = ( ) => { };`

assignments/function-conversion.js
```diff
@@ -1,23 +1,15 @@
...
 
-// let anotherFunction = function (param) {
-//   return param;
-// };
+let anotherFunction = param => param;
```
holly-g, line 5: Don't forget the `( )` around the first param!

PR #120, opened 2018-06-26
Carey Baldwin

careybaldwin02:


TERR1E (commented):
Hi Carey, great job with - the code looks clean and seems like you got a good grasp on most of the concepts we've gone over today!

Here is a link you may find helpful:

PR #124, opened 2018-06-26
Francis Tse - JavaScript II Assignment

francistse23:
Callbacks.js assignment completed. Have not changed functions to ES6 syntax


DixieKorley:
Got it. Thanks for the note in the description. 👍

PR #125, opened 2018-06-26
Joseph Stossmeister: Main Finished

Jstoss:
- function-conversion completed

### DIFF
--- a/assignments/array-methods.js
+++ b/assignments/array-methods.js
@@ -57,22 +57,22 @@ const runners = [{"id":1,"first_name":"Charmain","last_name":"Seiler","email":"c
 // The event director needs both the first and last names of each runner for their running bibs.  Combine both the first and last names into a new array called fullName.
 let fullName = [];
 
-runners.forEach (runner => { fullName.push(`${runner.first_name} ${runner.last_name}`); });
+runners.forEach (runner => fullName.push(`${runner.first_name} ${runner.last_name}`) );
 console.log(fullName);
 
 // ==== Challenge 2: Use .map() ====
 // The event director needs to have all the runner's first names converted to uppercase because the director BECAME DRUNK WITH POWER. Convert each first name into all caps and log the result
-let allCaps = runners.map(runner => { return runner.first_name.toUpperCase(); });
+let allCaps = runners.map(runner => runner.first_name.toUpperCase() );
 console.log(allCaps);
 
 // ==== Challenge 3: Use .filter() ====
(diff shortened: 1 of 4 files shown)


assignments/array-methods.js
```diff
@@ -3,81 +3,98 @@
...
 // ==== Challenge 3: Use .filter() ====
 // The large shirts won't be available for the event due to an ordering issue.  Get a list of runners with large sized shirts so they can choose a different size. Return an array named largeShirts that contains information about the runners that have a shirt size of L and log the result
-let largeShirts = [];
+let largeShirts = runners.filter( runner => runner.shirt_size == 'L' );
 console.log(largeShirts);
 
 // ==== Challenge 4: Use .reduce() ====
 // The donations need to be tallied up and reported for tax purposes. Add up all the donations into a ticketPriceTotal array and log the result
-let ticketPriceTotal = [];
+let ticketPriceTotal = runners.reduce( (donation, runner) => donation += runner.donation, 0 );
```
EricHech, line 75: I don't see a need to set the starting val to 0 here, just a heads up.

assignments/array-methods.js
```diff
@@ -3,81 +3,98 @@
...
+  if(!theBusinesses.includes(runner.company_name)) {
+    theBusinesses.push(runner.company_name);
+  }
+});
+theBusinesses.sort();
+console.log(theBusinesses);
 
-// Problem 3
\ No newline at end of file
+// Problem 3
```
EricHech, line 97: Awesome work in this file! Great creativity and commenting for your work.

For your filter, why not just do?:
`runner => runner.company_name !== 'Photospace')`

assignments/closure.js
```diff
@@ -1,19 +1,85 @@
 // ==== Challenge 1: Write your own closure ====
 // Write a simple closure of your own creation.  Keep it simple!
+let hundredWave = () => {
```
EricHech, line 3: Killin' it! Awesome example.

EricHech (commented):
You killed it man. Good job.

assignments/array-methods.js
Jstoss: I was having a weird bug that if I didn't set the starting value, it would freak out and return this long object value to the console. I meant to bring it up at the end of the day, but I completely forgot about it.

assignments/array-methods.js
Jstoss: I'm going to be completely honest. It's been so long since I've had to or need to use !== that I completely forgot that it existed.

PR #126, opened 2018-06-26
Naaz Shaikh

noshaikh:


Nate152:
Okay so some unfinished challenges but otherwise this all looks good. I know you feel that you aren't getting anywhere but as long as this is your code (which I know it is) we are making good progress. I can't find any glaring issues and the solutions that you do have look good. We can continue to go over why they work and what they are doing during each step the ensure you feel comfortable with it all. I know its a lot. Like we talked about, continue to put in the time and you will see results. Anyone can learn this stuff. Absolutely anyone.

PR #135, opened 2018-06-26
Adrian Bracamonte

adrianbraca:


assignments/closure.js
```diff
@@ -1,19 +1,48 @@
...
+closure()
 
 // ==== Challenge 2: Create a counter function ====
 const counter = () => {
   // Return a function that when invoked increments and returns a counter variable.
+  let count = 0
+  const increment = () => {
+    return ++count
+  }
+  return increment()
```
dtacheny, line 20: For this we want to return the function `increment` itself, not the result of invoking `increment()`. We want `newCounter` to be equivalent to `increment`.

So here's what we want to happen:

1. Invoke `counter()` and assign the result to `newCounter`

Right now, `newCounter` has a value of `1`, because that's the result of invoking `increment` on line 20. Instead, we want `newCounter` to have value that is the function `increment`. So we'd need to take the parentheses off of `increment()` on line 20.

2. We want to be able to invoke `newCounter` to increment the `count` variable inside of the `counter` function.

Since we now have `newCounter` assigned to the function `increment`, we can invoke it just like any other function. So on lines 26 and 27, we want to add parentheses after `newCounter` to invoke the `increment` function and increment the `count` variable.

dtacheny (commented):
Great work. It seems like you have a good understanding of these concepts!

PR #139, opened 2018-06-26
Charlie Winslow

cwinslow22:
Here for the pull request

### DIFF
--- a/assignments/closure.js
+++ b/assignments/closure.js
@@ -20,5 +20,17 @@ const counterFactory = () => {
   // Return an object that has two methods called `increment` and `decrement`.
   // `increment` should increment a counter variable in closure scope and return it.
   // `decrement` should decrement the counter variable and return it.
+  let count = 0
+  let object = {increment: () => {return ++count}, decrement: () => {return --count}};
+  // object.increment = () => {return ++count};
+  // object.decrement = () => {return --count};
+  return object;
   
 };
+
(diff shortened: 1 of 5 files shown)


assignments/closure.js
```diff
@@ -1,19 +1,36 @@
 // ==== Challenge 1: Write your own closure ====
 // Write a simple closure of your own creation.  Keep it simple!
-
+let simple = () => {let x = 5};
```
Nate152, line 4: I'm not sure how clearly this demonstrates that you understand closure. I can see how you could convince me, but without having that conversation this isn't super explicit. Josh's example was this:
```
let myName = "Josh";

function sayHello() {
  let example = "jim";
  // accessing myName outside this function
  return `Hello ${myName}`;
}
console.log(sayHello());

console.log(example);
```
Note how he defines the variable outside the scope of the function, but calls it from within. There are also variables defined within the function that we could talk about not being able to be called from outside. Just a bit more explicit in its demonstration of the concept.

Nate152:
Good job Charlie. I love to see the steady improvement over the course of the week. We will continue to cement the fundamentals as we go but I think you are doing a great job so far. This was day 2. Its all about perspective a lot of the time and it's super easy to lose sight of that. Especially when we are frustrated. Keep having a great attitude and we will keep making sure you have the tools to succeed.

PR #141, opened 2018-06-26
Grant Reighard

grantreighard:
- Working on function-conversion.js.

### DIFF
--- a/assignments/array-methods.js
+++ b/assignments/array-methods.js
@@ -1,4 +1,4 @@
-// A local community center is holding a fund rasising 5k fun run and has invited 50 small businesses to make a small donation on their behalf for some much needed updates to their facilities.  Each business has assigned a representative to attend the event along with a small donation.
+// A local community center is holding a fund raising 5k fun run and has invited 50 small businesses to make a small donation on their behalf for some much needed updates to their facilities.  Each business has assigned a representative to attend the event along with a small donation.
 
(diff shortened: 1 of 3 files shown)


assignments/array-methods.js
```diff
@@ -56,28 +56,47 @@ const runners = [{"id":1,"first_name":"Charmain","last_name":"Seiler","email":"c
...
 console.log(largeShirts);
 
 // ==== Challenge 4: Use .reduce() ====
 // The donations need to be tallied up and reported for tax purposes. Add up all the donations into a ticketPriceTotal array and log the result
 let ticketPriceTotal = [];
+ticketPriceTotal = runners.reduce((total, runner) => {return total += runner.donation}, 0);
 console.log(ticketPriceTotal);
 
 // ==== Challenge 5: Be Creative ====
 // Now that you have used .forEach(), .map(), .filter(), and .reduce().  I want you to think of potential problems you could solve given the data set and the 5k fun run theme.  Try to solve 3 unique problems using one or many of the array methods listed above.
```
Nate152, line 82: Ashwin pointed out the endsWith property for strings in JS. Another cool way you could have approached this same problem. Both feed the caveman.

Nate152:
Looking good. You clearly have a good grasp of the concepts so far. Like I said in the standup, I do have some more material that I can pass your way if you would like more practice. A lot of the stuff we did back in the day had a pretty high difficulty factor so it may be good practice for you. Let me know.

PR #143, opened 2018-06-26
David White conversion complete

kauaicrusier:


TERR1E (commented):
Seems like you were on the right track with function conversion.
Here are some helpful links on ES6 arrow syntax and array methods:

PR #144, opened 2018-06-26
Mike Kerbleski 

MKerbleski:


TERR1E (commented):
Hi Mike,

Here are some links you may find helpful:

MKerbleski:
Thank you Terrie!

assignments/array-methods.js
```diff
@@ -54,30 +54,77 @@ const runners = [{"id":1,"first_name":"Charmain","last_name":"Seiler","email":"c
...
+
+console.log(`\nGuests who ordered large t-shirts: \n${largeNames}`);
 
 // ==== Challenge 4: Use .reduce() ====
 // The donations need to be tallied up and reported for tax purposes. Add up all the donations into a ticketPriceTotal array and log the result
+
 let ticketPriceTotal = [];
-console.log(ticketPriceTotal);
+
+runners.forEach(function(x){ticketPriceTotal.push(x.donation)})
```
dtacheny, line 87: This works!

See if you can figure out how to combine lines 87 and 89 into a single `.reduce` 😉

assignments/closure.js
```diff
@@ -1,19 +1,108 @@
...
+
+// ==== Challenge 2: Create a counter function ====
+// Return a function that when invoked increments and returns a counter variable.
 // Example usage: const newCounter = counter();
 // newCounter(); // 1
 // newCounter(); // 2
 
+const cows = () => {
+  let count = 80;
+      return {
```
dtacheny, line 32: For this one we want to return a function, not an object.

So the function you have for the value of `add` in the returned object, you could just return that! 
```
return () => {
  count++;
  return count;
}
```
Then you would be able to invoke `cow()` on line 41 instead of `cow.add()`. This is because the value of `cow` will be the function returned after invoking `cows()`.

dtacheny (commented):
Everything looks good! Let me know if you have any questions/need further clarification on closure.

Nice work 👍

PR #145, opened 2018-06-26
Alex Caves

AlexanderCaves:


TERR1E (commented):
Hi Alex,

I only see the initial commit in your PR. Are there any specific blockers we can go over with together?

Here some links you may find useful to give another attempt at the assignment:

I look forward to seeing some progress and will talk to you tomorrow. 👍

AlexanderCaves:
Thanks for this Terrie, it all looks very helpful.  My plan is to go
through these links, the WTF is THIS video that Josh mentioned, and the
refresher that Michael is organizing in our group channel.  Then I'm going
to go back through all the projects starting with Monday and redo them.  I
feel really good about today's lecture so I just need to dig myself out a
bit.  I'll push commits on all three projects as I go, so if you're
interested feel free to check in on my progress.
Alex

PR #147, opened 2018-06-26
Keith Kennedy 

Kennedykid1995:


TERR1E (commented):
Hi Keith, it looks like you are steadily on your way of grasping the materials from the past two days.

Below are some links you may find helpful:

Keep up the good work 👍

PR #149, opened 2018-06-26
James Phillips

CASDFJ:
Callbacks done


assignments/array-methods.js
```diff
@@ -56,28 +56,77 @@ const runners = [{"id":1,"first_name":"Charmain","last_name":"Seiler","email":"c
 // ==== Challenge 1: Use .forEach() ====
 // The event director needs both the first and last names of each runner for their running bibs.  Combine both the first and last names into a new array called fullName. 
 let fullName = [];
+runners.forEach(function(element){fullName.push(element.first_name + ' ' + element.last_name)});
```
EricHech, line 59: The second version is styled more preferably, though with arrow syntax you could stick this all on one line. Also, this is a good use-case for string interpolation and template literals (backticks and dollar-sign curly braces).

assignments/array-methods.js
```diff
@@ -56,28 +56,77 @@ const runners = [{"id":1,"first_name":"Charmain","last_name":"Seiler","email":"c
...
 
 // ==== Challenge 2: Use .map() ====
 // The event director needs to have all the runner's first names converted to uppercase because the director BECAME DRUNK WITH POWER. Convert each first name into all caps and log the result
-let allCaps = [];
+let allCaps = runners.map(function(element) {return element['first_name'].toUpperCase()});
 console.log(allCaps); 
 
+//or//
+
+let allCaps = runners.map((person) => {
```
EricHech, line 80: Nice, this second one here looks really good.

assignments/array-methods.js
```diff
@@ -56,28 +56,77 @@ const runners = [{"id":1,"first_name":"Charmain","last_name":"Seiler","email":"c
...
 // Now that you have used .forEach(), .map(), .filter(), and .reduce().  I want you to think of potential problems you could solve given the data set and the 5k fun run theme.  Try to solve 3 unique problems using one or many of the array methods listed above.
 
 // Problem 1
+let email = [];
+runners.forEach(function(element){fullName.push(element.email)});
+console.log(email);
 
 // Problem 2
+let ids = [];
+ids = runners.filter(function(element){
```
EricHech, line 126: I don't think this `.filter()` is doing what you think it should. This is looking at every element in the array and checking to see if it has an `id` property, and if it does, then it will return that element. Therefore, since all elements in your array have `id`s, this filter doesn't change anything. Maybe if you did `return element.donation > 100`, then you could get an array of all the runners who donated more than $100. This is actually a practical use-case; maybe you want to put those runners in a unique category to recieve a special "thank you" email, for example.

assignments/callbacks.js
```diff
@@ -1,28 +1,41 @@
...
 
 function last(arr, cb) {
   // last passes the last item of the array into the callback.
+  cb(arr[arr.length - 1]);
 }
 
 function sumNums(x, y, cb) {
   // sumNums adds two numbers (x, y) and passes the result to the callback.
+  let sum = x + y;
+  cb(sum);
```
EricHech, line 22: This is good. You could also have done:
`cb(x + y)`

assignments/callbacks.js
```diff
@@ -1,28 +1,41 @@
...
 function multiplyNums(x, y, cb) {
   // multiplyNums multiplies two numbers and passes the result to the callback.
+  let product = x * y;
+  cb(product);
 }
 
 function contains(item, list, cb) {
   // contains checks if an item is present inside of the given array/list.
   // Pass true to the callback if it is, otherwise pass false.
+  if(list.includes(item) === true) {
```
EricHech, line 37: This works too:
`if(list.includes(item) === true) cb(true);`
`else cb(false);`

When one-lined, you don't need brackets, and the `else` will pair with the `if` directly above it. I would only do this for very short statements like you have here, though, and some linters don't like you to do it in the way that I'm showing you anyway because it makes it harder to edit the code by adding in more lines. Does that make sense?

assignments/closure.js
```diff
@@ -1,19 +1,45 @@
...
+function sayHello() {
+  let example = 'jim';
+  return `Hello ${myName}`;
+}
+
+console.log(sayHello());
+console.log(example);
 
 // ==== Challenge 2: Create a counter function ====
 const counter = () => {
```
EricHech, line 15: Did you grab this from the solution or from a fellow student, or did you write it yourself? Either is fine, it's just good for me to know.

assignments/closure.js
```diff
@@ -1,19 +1,45 @@
...
+newCounter() // 1
+newCounter() // 2
+newCounter() //3
 
 
 // ==== Challenge 3: Create a counter function with an object that can increment and decrement ====
 const counterFactory = () => {
+  let count = 0;
+  return {
+    increment: () => (++count),
```
EricHech, line 33: This is fine:
`increment: () => ++count,`

assignments/function-conversion.js
```diff
@@ -1,23 +1,56 @@
...
+let anotherFunction = (param) => {
+  return `${param}`;
+}
+
+//or
+
+let anotherFunction = param => `${param}`;
+
+//correct way//////////////////////////
+let anotherFunction = (param) => param;
```
EricHech, line 18: Good, so there is no need for the string interpolators here, and it seems like you realized that and removed them in the last example.

EricHech (commented):
Good stuff man! Keep up the good work. I hope my comments are helpful.

Also, awesome `commit` frequency!

PR #156, opened 2018-06-26
Rodrigo Acosta-Ramos

ghost:
initial commit


assignments/array-methods.js
```diff
@@ -56,28 +56,63 @@ const runners = [{"id":1,"first_name":"Charmain","last_name":"Seiler","email":"c
 // ==== Challenge 1: Use .forEach() ====
 // The event director needs both the first and last names of each runner for their running bibs.  Combine both the first and last names into a new array called fullName. 
 let fullName = [];
+fullName = runners.forEach(function(name) {
```
Nate152, line 59: Nice use of string interpolation and template literals here. Third different way I've seen to solve this problem so kudos for having an original approach.

assignments/closure.js
```diff
@@ -1,19 +1,39 @@
 // ==== Challenge 1: Write your own closure ====
 // Write a simple closure of your own creation.  Keep it simple!
+function rapperString() {
+  return(`lil ${string} `);
+}
+let string = 'bunny';
+console.log(rapperString());
```
Nate152, line 8: I think I know what you were getting at with this but I don't know that its the most explicit or simple example. I'd need to have you explain it to me to ensure that you really understand the concept, which I expect you do. My point is that a different example might more clearly show your understanding of the idea. This was the solution example:
```
let myName = "Josh";

function sayHello() {
  let example = "jim";
  // accessing myName outside this function
  return `Hello ${myName}`;
}
console.log(sayHello());

console.log(example);
```

Nate152:
Looks great. I love that you went back through and refactored all of your function syntax into arrow functions. Just good practice to help cement the idea. We will talk more about the downsides of arrow functions when we talk about the 'this' keyword so just keep in mind that there are downsides. Im glad that you are finding JS much more enjoyable.

PR #164, opened 2018-06-26
Stephanie Banks CS12

steviebanko:


assignments/closure.js
```diff
@@ -1,11 +1,27 @@
 // ==== Challenge 1: Write your own closure ====
 // Write a simple closure of your own creation.  Keep it simple!
 
+function styleFont(size, /*color*/){
```
Nate152, line 4: Like we talked about, a more explicit and easier explanation will typically involve two functions so there is a very clear use and need for the closure. Just keep this in mind as closure is a very common interview topic.

Nate152:
Awesome job Steph. This looks great. Try to be consistent with your use of either arrow functions or normal function syntax. Its not a huge deal in this particular case but we want to build good habits and a good habit is to be consistent with things of this nature. Another good example would be single versus double quotes. You can use either, you just need to be consistent whichever you choose. A little bit more that we can come back to and work on but I love to see that you put in some serious work here. Be proud and keep plugging away.

PR #170, opened 2018-06-26
working arrays

rverdi642:
Richard Verdier

### DIFF
--- a/assignments/closure.js
+++ b/assignments/closure.js
@@ -1,5 +1,15 @@
 // ==== Challenge 1: Write your own closure ====
 // Write a simple closure of your own creation.  Keep it simple!
+function counter(){
+	let counter = 0;
+} return{ function() {
+	counter = counter + 1;
+	conole.log ("in Anonymous function");
+	return counter;
+}
+}
+
+counter();
 
 
 // ==== Challenge 2: Create a counter function ====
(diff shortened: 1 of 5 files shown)


TERR1E (commented):
Hi Richard, great progress so far 👍 
Here are some helpful links for ES6 arrow syntax and closure:

assignments/closure.js
```diff
@@ -1,5 +1,15 @@
...
+function counter(){
+	let counter = 0;
+} return{ function() {
+	counter = counter + 1;
+	conole.log ("in Anonymous function");
+	return counter;
+}
+}
+
+counter();
```
dtacheny, line 12: There are a few small errors in this block. The logic is correct!

Line 5: That first curly brace shouldn't be there. It's closing off the `counter` function, so the `return` is cut out. Also, the return statement shouldn't wrapped in curly braces since that's being interpreted as an Object.

On line 12, you're invoking the counter function, but you aren't assigning the result to anything. The goal is to be able to invoke the function that's returned as a result of invoking `counter()`, so you could do something like this:
```
const newCounter = counter();
```
Since the `counter` function is returning an anonymous function, this line is effectively the same as this:
```
const newCounter = function() {
  counter = counter + 1;
  return counter;
}
```
but because of closure, `newCounter` will have access to the `counter` variable inside of the function! So then you can invoke `newCounter` and it will increment the `counter` variable.
```
console.log(newCounter()) // will print out 1
console.log(newCounter()) // will print out 2
```

I would also recommend naming the `counter` variable something different from the name of the function, for clarity.

dtacheny (commented):
Nice work! Let me know if you have any questions/need clarification about .filter or closure.

PR #177, opened 2018-06-26
learning foreach

bwhite06:


assignments/array-methods.js
```diff
@@ -54,23 +54,60 @@ const runners = [{"id":1,"first_name":"Charmain","last_name":"Seiler","email":"c
...
+// The event director needs both the first and last names of each runner for their running bibs.  Combine both the first and last names into a new array called fullName.
 let fullName = [];
+findrunner();
 console.log(fullName);
+function findrunner(){
+  runners.forEach(function(i) {
+    fullName.push(i.first_name);
+    fullName.push(i.last_name);
+
+  });
```
TERR1E: You did not have to put `.forEach` inside a function - just needed to use `.forEach` with a function to push `first_name` and `last_name` to fullName array

assignments/array-methods.js
```diff
@@ -54,23 +54,60 @@ const runners = [{"id":1,"first_name":"Charmain","last_name":"Seiler","email":"c
...
+  JSON.stringify(fullName);
```
TERR1E: You can just use a console.log which you used in line 60.

assignments/array-methods.js
```diff
@@ -54,23 +54,60 @@ const runners = [{"id":1,"first_name":"Charmain","last_name":"Seiler","email":"c
...
+
+
+
 // ==== Challenge 4: Use .reduce() ====
 // The donations need to be tallied up and reported for tax purposes. Add up all the donations into a ticketPriceTotal array and log the result
-let ticketPriceTotal = [];
+
+let ticketPriceTotal = 0;
+let result=[];
+for (var i = 0; i < runners.length; i++) {
```
TERR1E, line 104: Try not to use `var`

TERR1E (commented):
You are getting there - keep it up!
Here are some helpful links you can check out:

assignments/array-methods.js
groov1234: For your push you can also use this syntax to combine the two. ```variable.push(param.property + " " + param.property);```

PR #178, opened 2018-08-07
Michael Agard

fellanonymous:
@kait_schorr

### DIFF
--- a/assignments/callbacks.js
+++ b/assignments/callbacks.js
@@ -1,63 +1,107 @@
+/* PEER REVIEW NOTE: I cannot for the life of me follow the logic on the solution
+for the stretch goal. I've placed it at the bottom commented out for your convenience. */
+
+
 const items = ['Pencil', 'Notebook', 'yo-yo', 'Gum'];
 
+/* Problem 1
+firstItem passes the first item of the given array to the callback function. */
+
 function firstItem(arr, cb) {
-  // firstItem passes the first item of the given array to the callback function.
-  return cb(arr);
+  console.log(cb(arr))
 }
-function first(items) {
-  return console.log(items[0]);
+let first = function(items) {
+  return items[0];
 }
 firstItem(items, first)
 
 
-function getLength(arr, cb) {
-  // getLength passes the length of the array into the callback.
-  return cb(arr);
-}
-function length(items) {
-  return console.log(items.length);
-}
-firstItem(items, length);
+/* Problem #2
+getLength passes the length of the array into the callback. */
 
+function getLength(arr, cb) {
+  console.log(cb(arr));
+}
+let arrayLength = function(arr) {
+  return arr.length;
+}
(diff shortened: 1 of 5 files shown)


kait-schorr:
## Objectives
- [x] Student should be able to describe what closure is, how closure is created in a program and why it is important to have different types of `scope`
- [x] Student should be able to demonstrate the proper use a of a callback function by passing a function to another function as a parameter
- [x] Student should be able to demonstrate ability to use `.map` `.reduce` `.filter` and describe their use cases


## Great
* Your code is well structured.
* Good job committing more often!
* Your variable names are short and descriptive of what they represent, very nice!

## Requested Improvements
* Be sure you understand ES6 syntax and use it as frequently as possible (let, const, arrow functions, etc)
* Your section on callbacks is a little off. See line by line comments for clarification. 
* Your section on closures does not execute without errors.

## Questions
* What is the advantage of using ForEach or built-in array methods over for loops? The disadvantage?

## Rating: {1-3}
## Incomplete

assignments/callbacks.js
```diff
@@ -2,23 +2,52 @@ const items = ['Pencil', 'Notebook', 'yo-yo', 'Gum'];
 
 function firstItem(arr, cb) {
   // firstItem passes the first item of the given array to the callback function.
+  return cb(arr);
 }
+function first(items) {
+  return console.log(items[0]);
+}
+firstItem(items, first)
+
```
kait-schorr, line 7: So the instructions ask that `firstItem` pass the first item of its array into the callback function given to it. In practice this would look more like this:

 ```
function firstItem(arr, cb) {
  return cb(arr[0]);
}
```
For this exercise there isn't a specified callback function, but a similar correction needs to happen on all the future exercises as well. The way you have it structured now, the callback executing functions make no change and serve no purpose. Does that make sense? You want the logic to be inside of firstItem, not in a separate function. So, for example, you could have a simple print function. Something like: 
```
function print(output){
   console.log(output);
}
```
That you pass in as you "cb" for all the exercises. If you did this for first item, then your result would be to print the first item in the array you pass in. You'd invoke it like this:
```
firstItem(items, print);
```
And your output would be: `Pencil`

Try to go back through and restructure your callbacks. If you are still confused, let me know!

fellanonymous: If my brain hasn't melted too far out of my head at the end of today, I'll definitely take a look over my callbacks. Sadly, this is often the case and I find the weekend a better time to review my previous work. Here's to hoping today goes much better!

fellanonymous:
@kait-schorr 
## Questions
* What is the advantage of using ForEach or built-in array methods over for loops? The disadvantage?

Sorry it took me so long to answer this question. ForEach is often perfected since there are less areas of human error, and in return the code is much dryer.
The only disadvantage I've heard was that it uses more processing power, but not by much.

PR #179, opened 2018-08-07
Initial PR

verydecent:


gooseandmegander:
Thanks for the PR.

assignments/callbacks.js
```diff
@@ -1,30 +1,88 @@
...
 }
 
+firstItem(items, firstIndex);
+
+//
+
 function getLength(arr, cb) {
-  // getLength passes the length of the array into the callback.
+    // getLength passes the length of the array into the callback.
+  cb(arr);
```
gooseandmegander, line 20: If you don't return this value, it will log out as undefined.

assignments/callbacks.js
```diff
@@ -1,30 +1,88 @@
...
 }
 
+getLength(items, arrLength);
+
+//
+
 function last(arr, cb) {
-  // last passes the last item of the array into the callback.
+    // last passes the last item of the array into the callback.
+  cb(lastIndex);
```
gooseandmegander, line 33: `lastIndex` is a function, and you are not using the argument `items` given to last. What do you think is going on, and how can you fix it?

assignments/callbacks.js
```diff
@@ -1,30 +1,88 @@
...
+  return items[items.length - 1];
+}
+
+last(items, lastIndex);
+
+//
+
 function sumNums(x, y, cb) {
   // sumNums adds two numbers (x, y) and passes the result to the callback.
+  cb(x, y);
```
gooseandmegander, line 46: There's something you need to do here to get the sum to print or return to the console when you run the file. What needs to happen?

assignments/callbacks.js
```diff
@@ -1,30 +1,88 @@
...
 
+function listChecker(item, list) {
+  for(let i = 0; i < list.length; i++) {
+    if(item === list[i]) {
+      return true;
+    }
+  }
+}
+
+contains('Pencil', items, listChecker);
```
gooseandmegander, line 84: You're currently checking if an item _exists_ in the array and returning true if it does, but you're not returning false if the item does not exist.

assignments/array-methods.js
```diff
@@ -55,29 +55,86 @@ const runners = [{"id":1,"first_name":"Charmain","last_name":"Seiler","email":"c
 
 // ==== Challenge 1: Use .forEach() ====
 // The event director needs both the first and last names of each runner for their running bibs.  Combine both the first and last names into a new array called fullName. 
-let fullName = [];
+let fullName = runners.forEach(function(element, index) {
```
gooseandmegander, line 58: `forEach` returns undefined, which is why when you console.log fullName, it comes returns an array of undefined items. How would you mitigate this?

assignments/array-methods.js
```diff
@@ -55,29 +55,86 @@ const runners = [{"id":1,"first_name":"Charmain","last_name":"Seiler","email":"c
...
+});
 console.log(fullName);
 
+
+
 // ==== Challenge 2: Use .map() ====
 // The event director needs to have all the runner's first names converted to uppercase because the director BECAME DRUNK WITH POWER. Convert each first name into all caps and log the result
-let allCaps = [];
+let allCaps = runners.map(function(x) {
+    x.first_name.toUpperCase();
```
gooseandmegander, line 68: As is, this exercise is not logging out to the console. Why do you think that is?

assignments/array-methods.js
```diff
@@ -55,29 +55,86 @@ const runners = [{"id":1,"first_name":"Charmain","last_name":"Seiler","email":"c
...
+});
 console.log(allCaps); 
 
+
+
 // ==== Challenge 3: Use .filter() ====
 // The large shirts won't be available for the event due to an ordering issue.  Get a list of runners with large sized shirts so they can choose a different size. Return an array named largeShirts that contains information about the runners that have a shirt size of L and log the result
-let largeShirts = [];
+let largeShirts = runners.filter(function(x) {
+    if(x.shirt_size === "L";
```
gooseandmegander, line 77: As is, this exercise is not logging out to the console. Why do you think that is?

assignments/array-methods.js
```diff
@@ -55,29 +55,86 @@ const runners = [{"id":1,"first_name":"Charmain","last_name":"Seiler","email":"c
...
+});
 console.log(largeShirts);
 
+
+
 // ==== Challenge 4: Use .reduce() ====
 // The donations need to be tallied up and reported for tax purposes. Add up all the donations into a ticketPriceTotal array and log the result
-let ticketPriceTotal = [];
+let ticketPriceTotal = runners.reduce((reducer, runner) => {
+    return reducer += runner.donation;
```
gooseandmegander, line 86: No need for `+=`, `+` works here because the reducer method is already aggregating the results of the return.

gooseandmegander (commented):
Thanks, Wonjae! I've left some comments for you to review.

PR #180, opened 2018-08-07
Jordan's PR

blokboy:
everything without stretch


crypto-jones:
Overall, you code is clean & readable. A couple of suggestions:

1. Commit more often.
2. I'd recommend reviewing the lecture video from Tuesday to solidify the concepts. In your `array-methods.js`, you're not using some of the methods in the most efficient way. Also, I noticed you didn't use `.reduce` at all. It's super important to understand these array methods going forward, so definitely take some time to review.

PR #189, opened 2018-08-07
Javascript II challenges - Alejandro Krasovsky

alejandrok93:


crypto-jones:
Your code is clean, easy to read, and it all functions as expected. I love how you are committing multiple times and explicating what you did in the commit message. This is a useful habit to get into. Great job! 

There are just a few minor things I noticed:

1. In `array-methods.js` for challenges 2 - 4, you don't need to keep the empty array, e.g., you can clean it up by using `let allCaps = runners.map...` and just delete the line above it. These array methods automatically output a new array as the variable name with the desired data.

2. Now that you know about template literals, I'd recommend using them vs. concatenating multiple lines in a return statement or console.log.

Excellent job!

PR #190, opened 2018-08-07
done tasks 1 - 4

decagondev:


decagondev:
will have a quick coffee then take a look at some of the stretch goals

PR #196, opened 2018-08-07
Kevin Sooter - JS-II

kevin-sooter:


Velsu (commented):
Great job on the assignment Kevin. You successfully completed the MVP and even tackled some of the stretch goals. Your code is clean and some solutions are pretty clever.

The ony issue i was able to find was in the `callbacks` exercise with `indexOf` that is called on a, that one can cause issues and throw errors that it is not a function.

Your git workflow could be a little more descriptive with more frequent commit messages.

PR #208, opened 2018-08-07
Max Kajiwara - JavaScript-II

maxkajiwara:


crypto-jones:
Your code is clean, easy to read, and all functions as expected. Great job! I love how you are committing multiple times and explicating what you did in the commit message. 

You clearly have a solid grasp of the content. The only minor thing I noticed is in challenge 2 of `closure.js`. You would need to increment before and not after `count` if you want the accumulator  to increment to one for the first invocation of the function. Other than that, outstanding job on this assignment!

PR #211, opened 2018-08-07
JavaScript-II assignment

ckirkwood94:


gooseandmegander:
Thanks for the PR.

assignments/array-methods.js
```diff
@@ -56,28 +56,63 @@ const runners = [{"id":1,"first_name":"Charmain","last_name":"Seiler","email":"c
 // ==== Challenge 1: Use .forEach() ====
 // The event director needs both the first and last names of each runner for their running bibs.  Combine both the first and last names into a new array called fullName. 
 let fullName = [];
+runners.forEach(function(element) {
+  fullName.push(`${element.first_name} ${element.last_name}`);
```
gooseandmegander, line 60: Nice use of template literals!

assignments/array-methods.js
```diff
@@ -56,28 +56,63 @@ const runners = [{"id":1,"first_name":"Charmain","last_name":"Seiler","email":"c
...
+  return element.shirt_size === "L"; 
+});
+
 console.log(largeShirts);
 
 // ==== Challenge 4: Use .reduce() ====
 // The donations need to be tallied up and reported for tax purposes. Add up all the donations into a ticketPriceTotal array and log the result
-let ticketPriceTotal = [];
+let ticketPriceTotal = runners.reduce((theReducer, element) =>{
+  return theReducer += element.donation;
```
gooseandmegander, line 83: No need for `+=`, `+` works here because the reducer method is already aggregating the results of the return.

assignments/closure.js
```diff
@@ -1,11 +1,34 @@
...
 // Write a simple closure of your own creation.  Keep it simple!
+function sayName() {
+  let name = "Megan";
 
+  function greeting() {
+    let myName = "Caleb";
+    console.log(`Hey ${name}`);
+
+    function greetingFrom() {
+      console.log(`This is from ${myName}, in case you were wondering`);
```
gooseandmegander, line 11: Thanks Caleb :)

assignments/closure.js
```diff
@@ -1,11 +1,34 @@
...
+    }
+    greetingFrom();
+  }
+  greeting();
+}
+sayName();
 
 // ==== Challenge 2: Create a counter function ====
 const counter = () => {
```
gooseandmegander, line 20: This is how I tweaked yours to have closure:

>const counter = () => {
  let value = null;
  return function newCounter() {
    console.log((value += 1));
  };
};
const Acounter = counter();
console.log(Acounter());
console.log(Acounter());

The closure is created in Acounter, not counter(). Counter creates and destroys its scope whenever it is called. The Acounter needs to be set equal to the returned counter function outside of the counter function scope.

gooseandmegander (commented):
Everything looks great, Caleb. I have some comments for you to review. Go for stretch!

PR #217, opened 2018-08-07
Joseph Chretien

JGolden25:
@kait_schorr

### DIFF
--- a/assignments/callbacks.js
+++ b/assignments/callbacks.js
@@ -1,29 +1,50 @@
 const items = ['Pencil', 'Notebook', 'yo-yo', 'Gum'];
 
+function print (output){
+  console.log(output);
+}
+
+
 function firstItem(arr, cb) {
   // firstItem passes the first item of the given array to the callback function.
+ return cb(arr[0]);
 }
 
 function getLength(arr, cb) {
   // getLength passes the length of the array into the callback.
+  return cb(arr.length);
 }
 
 function last(arr, cb) {
   // last passes the last item of the array into the callback.
+  return cb(arr[arr.length-1]);
 }
 
 function sumNums(x, y, cb) {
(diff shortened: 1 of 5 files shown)


kait-schorr:
## Objectives
- [x] Student should be able to describe what closure is, how closure is created in a program and why it is important to have different types of `scope`
- [x] Student should be able to demonstrate the proper use a of a callback function by passing a function to another function as a parameter
- [x] Student should be able to demonstrate ability to use `.map` `.reduce` `.filter` and describe their use cases

## Great
* The array-methods and callback exercises look great!
* Your code is well structured. 
* Good job attempting the stretch goal!

## Requested Improvements
* Try to commit more frequently and with meaningful messages
* The assignment is incomplete. You still need to do the following:
  * Challenge 5 in arrays, Write your own questions and solve them!
  * Finish the closure exercises
* See line by line comments for suggestions/corrections

## Questions
* What benefits do callbacks offer us?

## Rating: {1-3}
## Incomplete

assignments/callbacks.js
```diff
@@ -1,29 +1,50 @@
...
 function contains(item, list, cb) {
   // contains checks if an item is present inside of the given array/list.
   // Pass true to the callback if it is, otherwise pass false.
+for(let i = 0; i <  list.length; i++){
+if (list[i] === item){
+  return true;
+}
+else {
+  return false;
 }
```
kait-schorr, line 42: So this method will not actually work for contains. There are two possible scenarios:
1. The element you are looking for is the first element in the array, so your method returns true
2. The element you are looking for is either in the array elsewhere, or isn't in the array and  your method returns false

The reason for this is the structure of your if statement. Essentially as soon as you run the if statement one time, one of the returns will fire and the function will be exited. If you move the `return false` out of your for loop, it should fix the problem.

PR #222, opened 2018-08-07
Andrew Ghobrial - JavaScript-II

ghobs91:
@gooseandmegander

### DIFF
--- a/assignments/array-methods.js
+++ b/assignments/array-methods.js
@@ -56,18 +56,21 @@ const runners = [{"id":1,"first_name":"Charmain","last_name":"Seiler","email":"c
 // ==== Challenge 1: Use .forEach() ====
 // The event director needs both the first and last names of each runner for their running bibs.  Combine both the first and last names into a new array called fullName. 
 let fullName = [];
-
 runners.forEach(function(element) {
     let runnerFullName = (element.first_name + element.last_name);
     fullName.push(runnerFullName);
 });
-
 console.log(fullName);
-
 
 // ==== Challenge 2: Use .map() ====
 // The event director needs to have all the runner's first names converted to uppercase because the director BECAME DRUNK WITH POWER. Convert each first name into all caps and log the result
 let allCaps = [];
+
+runners.forEach(function(element) {
+    let capitalizedName = (element.first_name.toUpperCase());
+    allCaps.push(capitalizedName);
+});
+
 console.log(allCaps); 
 
 // ==== Challenge 3: Use .filter() ====
--- a/assignments/array-methods.js
+++ b/assignments/array-methods.js
@@ -65,17 +65,21 @@ console.log(fullName);
 // ==== Challenge 2: Use .map() ====
 // The event director needs to have all the runner's first names converted to uppercase because the director BECAME DRUNK WITH POWER. Convert each first name into all caps and log the result
 let allCaps = [];
-
 runners.forEach(function(element) {
     let capitalizedName = (element.first_name.toUpperCase());
     allCaps.push(capitalizedName);
 });
-
 console.log(allCaps); 
 
 // ==== Challenge 3: Use .filter() ====
 // The large shirts won't be available for the event due to an ordering issue.  Get a list of runners with large sized shirts so they can choose a different size. Return an array named largeShirts that contains information about the runners that have a shirt size of L and log the result
 let largeShirts = [];
+const largeShirtsFilter = runners.filter((data) =>{
+    if (data.shirt_size == 'L') {
+        largeShirts.push (data);
+    }
+});
+return largeShirtsFilter;
 console.log(largeShirts);
 
 // ==== Challenge 4: Use .reduce() ====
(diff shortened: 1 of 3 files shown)


gooseandmegander:
Thanks for the PR.

assignments/callbacks.js
```diff
@@ -2,27 +2,43 @@ const items = ['Pencil', 'Notebook', 'yo-yo', 'Gum'];
 
 function firstItem(arr, cb) {
   // firstItem passes the first item of the given array to the callback function.
+  return cb(arr[0]);
```
gooseandmegander, line 5: This is correct, and to check yourself, you need to invoke firstItem and create a callback function to pass as an argument to the cb parameter. For instance:
>function firstItem(arr, cb) {
  return cb(arr[0]);
}

>let first = function(item) {
  console.log(item);
};

>firstItem(items, first);

assignments/callbacks.js
```diff
@@ -2,27 +2,43 @@ const items = ['Pencil', 'Notebook', 'yo-yo', 'Gum'];
...
 
 function getLength(arr, cb) {
   // getLength passes the length of the array into the callback.
+  return cb(arr.length);
 }
 
+
 function last(arr, cb) {
   // last passes the last item of the array into the callback.
+  let lastItemIndex = arr.length;
```
gooseandmegander, line 16: `arr.length` will give the amount of items in the array, but we want the index of the last item. Think about how array indices increment.

assignments/callbacks.js
```diff
@@ -2,27 +2,43 @@ const items = ['Pencil', 'Notebook', 'yo-yo', 'Gum'];
...
+  return cb(sumResult);
 }
 
 function multiplyNums(x, y, cb) {
   // multiplyNums multiplies two numbers and passes the result to the callback.
+  let multiplyResult = (x * y);
+  return cb(multiplyResult);
 }
 
 function contains(item, list, cb) {
```
gooseandmegander, line 32: Nice use of the flag and conditional.

assignments/array-methods.js
```diff
@@ -56,22 +56,40 @@ const runners = [{"id":1,"first_name":"Charmain","last_name":"Seiler","email":"c
 // ==== Challenge 1: Use .forEach() ====
 // The event director needs both the first and last names of each runner for their running bibs.  Combine both the first and last names into a new array called fullName. 
 let fullName = [];
+runners.forEach(function(element) {
+    let runnerFullName = (element.first_name + element.last_name);
```
gooseandmegander, line 60: Don't forget the space between the names.

assignments/array-methods.js
```diff
@@ -56,22 +56,40 @@ const runners = [{"id":1,"first_name":"Charmain","last_name":"Seiler","email":"c
...
 
 // ==== Challenge 3: Use .filter() ====
 // The large shirts won't be available for the event due to an ordering issue.  Get a list of runners with large sized shirts so they can choose a different size. Return an array named largeShirts that contains information about the runners that have a shirt size of L and log the result
 let largeShirts = [];
+const largeShirtsFilter = runners.filter((data) =>{
+    if (data.shirt_size == 'L') {
+        largeShirts.push (data);
+    }
+});
+return largeShirtsFilter;
```
gooseandmegander, line 82: This return is outside of function scope. If you replace `return largeShirtsFilter;` with simple `largeShirtsFilter;`, the filtered results will log onto the console. You haven't actually called largeShirtsFilter otherwise.

gooseandmegander: `filter` also creates a new array on its own, so you don't need to push each data piece into the `largeShirts` array. Logging out the newly created array stored in `largeShirtsFilter` and tweaking the filter method by returning the conditional `data.shirt_size == 'L'` will do the trick. `console.log(largeShirtsFilter;`

assignments/array-methods.js
```diff
@@ -56,22 +56,40 @@ const runners = [{"id":1,"first_name":"Charmain","last_name":"Seiler","email":"c
...
 console.log(largeShirts);
 
 // ==== Challenge 4: Use .reduce() ====
 // The donations need to be tallied up and reported for tax purposes. Add up all the donations into a ticketPriceTotal array and log the result
-let ticketPriceTotal = [];
-console.log(ticketPriceTotal);
+
+const ticketPriceTotal = runners.reduce((theReducer, item) =>{
+    return theReducer += item.donation;
```
gooseandmegander, line 89: No need for `+=`, `+` works here because the reducer method is already aggregating the results of the return.

assignments/closure.js
```diff
@@ -1,15 +1,31 @@
 // ==== Challenge 1: Write your own closure ====
 // Write a simple closure of your own creation.  Keep it simple!
-
+function sayCar() {
+  let favoriteCar = "Tesla Model S";
```
gooseandmegander, line 4: Teslas are so cool.

assignments/closure.js
```diff
@@ -1,15 +1,31 @@
...
+  function whyLove() {
+    console.log(favoriteCar + " is my favorite because it's the FUTURE");
+  }
+  whyLove();
+}
+sayCar();
 
 // ==== Challenge 2: Create a counter function ====
+let counterVariable = 0;
```
gooseandmegander, line 13: The counterVariable should be inside the function or it isn't closure. : P

gooseandmegander (commented):
Nice work, Andrew. I've got some notes for you to review.

For callbacks.js, add in the function invocations and any callback functions necessary to make the exercises work. I gave you one example.

PR #237, opened 2018-08-08
Keiran Kozlowski

gooseandmegander:
Creating PR to make code review comments.

### DIFF
--- a/assignments/array-methods.js
+++ b/assignments/array-methods.js
@@ -56,21 +56,46 @@ const runners = [{"id":1,"first_name":"Charmain","last_name":"Seiler","email":"c
 // ==== Challenge 1: Use .forEach() ====
 // The event director needs both the first and last names of each runner for their running bibs.  Combine both the first and last names into a new array called fullName. 
 let fullName = [];
+runners.forEach(function (runner) {
+    fullName.push(runner["first_name"] + " " + runner["last_name"]);
+});
+
 console.log(fullName);
 
 // ==== Challenge 2: Use .map() ====
 // The event director needs to have all the runner's first names converted to uppercase because the director BECAME DRUNK WITH POWER. Convert each first name into all caps and log the result
 let allCaps = [];
+let firstNames = [];
+
+runners.forEach(function (runner) {
+    firstNames.push(runner["first_name"]);
+});
+console.log(firstNames);
+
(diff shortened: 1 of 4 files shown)


assignments/callbacks.js
```diff
@@ -2,33 +2,70 @@ const items = ['Pencil', 'Notebook', 'yo-yo', 'Gum'];
...
+});
+
+
 function getLength(arr, cb) {
   // getLength passes the length of the array into the callback.
+    return cb(arr.length);
 }
 
+getLength(items, function (length) {
+    console.log(items.length);
```
gooseandmegander, line 19: You're not actually using the parameter `length` in the function's block. It would be more appropriate to use `console.log(length);`

assignments/array-methods.js
```diff
@@ -56,28 +56,130 @@ const runners = [{"id":1,"first_name":"Charmain","last_name":"Seiler","email":"c
...
+});
+
 console.log(fullName);
 
 // ==== Challenge 2: Use .map() ====
 // The event director needs to have all the runner's first names converted to uppercase because the director BECAME DRUNK WITH POWER. Convert each first name into all caps and log the result
 let allCaps = [];
+let firstNames = [];
+
+runners.forEach(function (runner) {
```
gooseandmegander, line 70: Do you think it is necessary to use both `forEach` and `map`? Why?

assignments/array-methods.js
```diff
@@ -56,28 +56,130 @@ const runners = [{"id":1,"first_name":"Charmain","last_name":"Seiler","email":"c
...
 // The large shirts won't be available for the event due to an ordering issue.  Get a list of runners with large sized shirts so they can choose a different size. Return an array named largeShirts that contains information about the runners that have a shirt size of L and log the result
 let largeShirts = [];
+largeShirts = runners.filter(runner => runner["shirt_size"] === "L");
+
 console.log(largeShirts);
 
 // ==== Challenge 4: Use .reduce() ====
 // The donations need to be tallied up and reported for tax purposes. Add up all the donations into a ticketPriceTotal array and log the result
 let ticketPriceTotal = [];
+let prices = [];
```
gooseandmegander, line 89: Why did you add the prices functionality in?

gooseandmegander (commented):
Nice job, Keiran. I've left some comments for you to review.

PR #253, opened 2018-08-16
Jerrard Smith JavaScript II WIP

JerrardSmithOfficial:
@47analogy


47analogy:
Nice work on this assignment. You have a good understanding of the material for this unit. Want to note that for the function last exercise in callbacks.js, you want to pass arr[arr.length - 1] in the callback, otherwise you'll just get just the index. Keep up the good work!

PR #268, opened 2018-08-17
Andres Rivera

andresriveratoro:
@47analogy


47analogy:
Nice work in completing the assignment, especially for doing the stretch goals!

In closure.js (Challenge 2), wanted to note that you can also return the count (instead of logging) to 
`  let count = 0;
  const addCount = () => {
    count++;
    console.log(count); // return here
  }
  return addCount;`

PR #289, opened 2018-08-17
Stefan Clem - JavaScript-II

iepoch:
@47analogy


47analogy:
Nice job on the assignment. Extra points for creating the callback function to log/check your code. Same for the stretch assignments. Good job of getting used to the ternary operator.  Code is clean, consistent, and there aren't any errors. Excellent work, especially on the stretch goals!

PR #296, opened 2018-08-17
Bruce Cabanayan - Javascript II

bcabanayan:
Submission for Javascript II project!

### DIFF
--- a/assignments/closure.js
+++ b/assignments/closure.js
@@ -1,14 +1,31 @@
 // ==== Challenge 1: Write your own closure ====
-// Write a simple closure of your own creation.  Keep it simple!
+const divisibleBy = (x) => {
+  return checkDivisibility = (y) => {
+    return (y % x === 0);
+  } 
+}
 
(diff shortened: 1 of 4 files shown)


assignments/closure.js
```diff
@@ -1,15 +1,32 @@
...
-// Write a simple closure of your own creation.  Keep it simple!
+const divisibleBy = (x) => {
+  return checkDivisibility = (y) => {
+    return (y % x === 0);
+  } 
+}
 
+let divByFive = divisibleBy(5);
+
+divByFive(100); //returns true
```
kiachoi, line 10: You can also call it as `divisibleBy(5)(100)` instead of calling them separately.

kiachoi (approved):
✔️ What Went Well:
----
+ Callbacks
+ Array Methods
+ Closures
+ Arrow Function Syntax
+ Nicely written code and frequent commit messages

🎗 Recommendation(s):
----
+ Aside from the calling closure functions - none!

Other:
----
+ 3/3

PR #299, opened 2018-08-17
Wes Meeks

ghost:
@kiadorii

### DIFF
--- a/assignments/array-methods.js
+++ b/assignments/array-methods.js
@@ -54,23 +54,32 @@ const runners = [{"id":1,"first_name":"Charmain","last_name":"Seiler","email":"c
 {"id":50,"first_name":"Shell","last_name":"Baine","email":"sbaine1d@intel.com","shirt_size":"M","company_name":"Gabtype","donation":171}];
 
 // ==== Challenge 1: Use .forEach() ====
-// The event director needs both the first and last names of each runner for their running bibs.  Combine both the first and last names into a new array called fullName. 
+// The event director needs both the first and last names of each runner for their running bibs.  Combine both the first and last names into a new array called fullName.
 let fullName = [];
+runners.forEach(function(item, index, array) {
+  return fullName.push(`${runners[index].first_name} ${runners[index].last_name}`);
+})
 console.log(fullName);
 
 // ==== Challenge 2: Use .map() ====
 // The event director needs to have all the runner's first names converted to uppercase because the director BECAME DRUNK WITH POWER. Convert each first name into all caps and log the result
 let allCaps = [];
-console.log(allCaps); 
+ allCaps = runners.map(runner => {
+   return runner.first_name.toUpperCase()
+ });
+console.log(allCaps);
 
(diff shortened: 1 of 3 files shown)


assignments/array-methods.js
```diff
@@ -54,23 +54,32 @@ const runners = [{"id":1,"first_name":"Charmain","last_name":"Seiler","email":"c
...
-console.log(allCaps); 
+ allCaps = runners.map(runner => {
+   return runner.first_name.toUpperCase()
+ });
+console.log(allCaps);
 
 // ==== Challenge 3: Use .filter() ====
 // The large shirts won't be available for the event due to an ordering issue.  Get a list of runners with large sized shirts so they can choose a different size. Return an array named largeShirts that contains information about the runners that have a shirt size of L and log the result
-let largeShirts = [];
+let largeShirts = runners.filter((shirt_size, i, runners) => runners[i].shirt_size === "L");
```
kiachoi, line 74: You would only need to pass in the runner in the argument.
`let largeShirts = runners.filter((runner) => { runner.shirt_size === 'L' });`

assignments/closure.js
```diff
@@ -1,14 +1,29 @@
 // ==== Challenge 1: Write your own closure ====
 // Write a simple closure of your own creation.  Keep it simple!
+function nameIs(firstName, lastName){
+  var intro= "Your name is ";
+  function fullName(){
+    return intro + firstName + " " + lastName;
+  }
+  return fullName ();
+}
+nameIs ("Slim", "Shady");
```
kiachoi, line 10: Hah!! I love this. 😂

assignments/closure.js
```diff
@@ -1,14 +1,29 @@
...
 
 
 // ==== Challenge 2: Create a counter function ====
 const counter = () => {
   // Return a function that when invoked increments and returns a counter variable.
-};
+
+let counter = 0;
+let plusOne = () = >{
```
kiachoi, line 18: When defining functions, you would want to use `const plusOne` rather than `let plusOne`. This is because functions generally will not be called again nor should they be altered.

assignments/callbacks.js
```diff
@@ -2,29 +2,69 @@ const items = ['Pencil', 'Notebook', 'yo-yo', 'Gum'];
 
 function firstItem(arr, cb) {
   // firstItem passes the first item of the given array to the callback function.
+return cb(arr);
```
kiachoi, line 5: This should be written as `return cb(arr[0]);` without the need for `function getFirstItem(items) {...}`. This applies throughout the rest of your code. I believe we talked about this and am not sure if you have updated it since then.

kiachoi (approved):
✔️ What Went Well:
----
+ Array Methods
+ Closures
+ Arrow Function Syntax implementation in certain parts.
+ Clean code and commit messages

🎗 Recommendation(s):
----
+ Please take another look at callbacks.js
+ Frequency of commit messages

Other:
----
+ 2/3

PR #300, opened 2018-08-17
Louis Li - JavaScript-II 

louislee7:
@kiadorii

### DIFF
--- a/assignments/callbacks.js
+++ b/assignments/callbacks.js
@@ -2,28 +2,37 @@ const items = ['Pencil', 'Notebook', 'yo-yo', 'Gum'];
 
 function firstItem(arr, cb) {
   // firstItem passes the first item of the given array to the callback function.
+  cb(arr[1]);
 }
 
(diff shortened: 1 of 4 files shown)


assignments/callbacks.js
```diff
@@ -2,28 +2,37 @@ const items = ['Pencil', 'Notebook', 'yo-yo', 'Gum'];
 
 function firstItem(arr, cb) {
   // firstItem passes the first item of the given array to the callback function.
+  cb(arr[0]);
```
kiachoi: Make sure to return this.

kiachoi (approved):
✔️ What Went Well:
----
+ Callbacks
+ Array Methods
+ Closures
+ Arrow function syntax
+ Clean commit messages and code

🎗 Recommendation(s):
----
+ When you feel more comfortable, start using `let` or `const` in replacement to `var`.

Other:
----
+ 3/3

PR #304, opened 2018-08-18
Humberto Raya Javascript-II

hraya:
@clarakosi

### DIFF
--- a/assignments/callbacks.js
+++ b/assignments/callbacks.js
@@ -2,7 +2,18 @@ const items = ['Pencil', 'Notebook', 'yo-yo', 'Gum'];
 
 function firstItem(arr, cb) {
   // firstItem passes the first item of the given array to the callback function.
+  for(let i = 0; i < arr.length; i++) {
+    if (arr[i] === arr[0]) {
+      cb(arr[0])
+    }
+  }
 }
+
+function logFirstItem (item) {
+  console.log(`${item} is the first item in the array!`);
+}
+
+firstItem(items, logFirstItem)
 
 function getLength(arr, cb) {
   // getLength passes the length of the array into the callback.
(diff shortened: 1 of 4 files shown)


assignments/callbacks.js
```diff
@@ -2,29 +2,88 @@ const items = ['Pencil', 'Notebook', 'yo-yo', 'Gum'];
 
 function firstItem(arr, cb) {
   // firstItem passes the first item of the given array to the callback function.
+  for(let i = 0; i < arr.length; i++) {
+    if (arr[i] === arr[0]) {
```
clarakosi, line 6: Since `i` is starting at 0 `arr[i]` will be equivalent to `arr[0]`. The first element in the array will always be at `arr[0]`. So there was no need for this for loop. You could just pass `arr[0]` to the callback.

assignments/callbacks.js
```diff
@@ -2,29 +2,88 @@ const items = ['Pencil', 'Notebook', 'yo-yo', 'Gum'];
...
+}
+
+function lastItem(item) {
+  console.log(`${item} is the last item of the array!`);
 }
 
+last(items, lastItem)
+
 function sumNums(x, y, cb) {
   // sumNums adds two numbers (x, y) and passes the result to the callback.
```
clarakosi, line 41: Great solution but the function should have passed the sum to the callback. The same for the other 2 problems below it.

assignments/closure.js
```diff
@@ -1,10 +1,27 @@
 // ==== Challenge 1: Write your own closure ====
 // Write a simple closure of your own creation.  Keep it simple!
+const greetingFunc = () => {
+  const nickname = "The Homie"
+
+  const introduceSelf = () => {
+    return `I am ${nickname} from the states! and you?`
+  }
+
+  console.log(introduceSelf());
```
clarakosi, line 10: Consider rewriting these with `return` instead of `console.log`

clarakosi (commented):
Great work, Humberto! 👍  

I did leave a few comments about how you could improve your code.

PR #308, opened 2018-08-18
Chris Juelfs - JavaScript-II: all assignments done!

CheapCharlie:
i couldnt figure out how to do the function in lines 46-56 of the callbacks subassignment without using .includes so if its possible please explain how. TYVM!

### DIFF
--- a/assignments/array-methods.js
+++ b/assignments/array-methods.js
@@ -55,29 +55,50 @@ const runners = [{"id":1,"first_name":"Charmain","last_name":"Seiler","email":"c
 
 // ==== Challenge 1: Use .forEach() ====
 // The event director needs both the first and last names of each runner for their running bibs.  Combine both the first and last names into a new array called fullName. 
+
 let fullName = [];
+
+runners.forEach(function(element) {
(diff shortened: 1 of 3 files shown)


47analogy:
Nice job on the assignment. Your work shows that you have a solid understanding of the concepts and how to implement them. Extra points for creating the callback function to log and test your code. 

Concerning your question, the includes method was created for this exact purpose and most modern browsers support it. You probably could also achieve the same results using indexOf(). 

In the arrays assignment, just wanted to note that you'll mainly see object properties used with dot notation, i.e. element.first_name vs. element["first_name"]

In the closures assignment, just wanted to note that in es6, const and let are better (use let when the variable can be reassigned). In Challenge 1, myName is defined in the global scope, so it can be seen by any function. With a closure, you want to keep it as a reference to the original function's scope. For Challenge 2, a possible solution is to replace line 17: console.log(++count); with return ++count; inside of the addOne function.

CheapCharlie:
any notes you have on when to use const vs: var vs: let would be greatly appreciated since i don't ever remember Dan specifically going over it.

PR #317, opened 2018-08-22
Mehdi: Completed Assignment

mabbashm110:
@analogy47 completed JS II


47analogy:
Excellent job of creating a list (array) of just the names in Challenge #3 in the arrays assignment. In Challenge #4 of the same assignment, you can use reduce to sum up the array, something like:  
ticketPriceTotal = runners.reduce((total, amount) => total + amount.donation, 0);

In callbacks assignment, for the function named last, it would be possible to pass the last item in the array to the callback:
cb(arr[arr.length - 1]); // arr.length - 1 would give the last element

Also, nice work on the closures assignment

PR #326, opened 2018-09-11
Chad Jemmett Javascript-II

chadjemmett:
@sumi419  ( o_o)-b


sumi419:
Good job Chad! I think you might need to resolve the conflicting branches here?

PR #328, opened 2018-09-11
Shawn Antonucci - JavaScript-II

shawnantonucci:
@kamry-bowman setup repo


assignments/array-methods.js
```diff
@@ -56,28 +56,67 @@ const runners = [{"id":1,"first_name":"Charmain","last_name":"Seiler","email":"c
...
 
 // ==== Challenge 3: Use .filter() ====
 // The large shirts won't be available for the event due to an ordering issue.  Get a list of runners with large sized shirts so they can choose a different size. Return an array named largeShirts that contains information about the runners that have a shirt size of L and log the result
 let largeShirts = [];
-console.log(largeShirts);
+largeShirts = runners.filter(function(runners) {
+    if(runners.shirt_size === "L") {
+        return runners.shirt_size;
+    }
+});
```
kamry-bowman, line 80: nice job here! maybe just use `return runners.shirt_size === L` instead. With what you have now, it returns 'L' if true, and undefined in false. Javascript coerces those into true and false, so it works, but it's better to be more explicit with booleans I think.

assignments/array-methods.js
```diff
@@ -56,28 +56,67 @@ const runners = [{"id":1,"first_name":"Charmain","last_name":"Seiler","email":"c
...
 let ticketPriceTotal = [];
+ticketPriceTotal = runners.reduce(function(total, runners) {
+    return total + runners.donation;
+}, 0);
 console.log(ticketPriceTotal);
 
 // ==== Challenge 5: Be Creative ====
 // Now that you have used .forEach(), .map(), .filter(), and .reduce().  I want you to think of potential problems you could solve given the data set and the 5k fun run theme.  Try to create and then solve 3 unique problems using one or many of the array methods listed above.
 
 // Problem 1
```
kamry-bowman, line 94: nice job on coming up with creative uses!

assignments/array-methods.js
```diff
@@ -56,28 +56,67 @@ const runners = [{"id":1,"first_name":"Charmain","last_name":"Seiler","email":"c
...
+
+// Problem 3
+// Gather info on the top donators.
 
-// Problem 3
\ No newline at end of file
+let highestDonations = [];
+highestDonations = runners.filter(function(runners) {
+    if(runners.donation > 290) {
+        return runners.donation;
```
kamry-bowman, line 119: you can get rid of the if block, and just have `return runners.donation > 290` on line 118

kamry-bowman (commented):
Great job, your code and creativity are both great!

PR #329, opened 2018-09-11
Jake Thomas

jmaxt12:


assignments/callbacks.js
```diff
@@ -20,25 +20,35 @@ const items = ['Pencil', 'Notebook', 'yo-yo', 'Gum'];
...
-
+// function cb(arg){
+//   console.log(arg)
+// }
 function getLength(arr, cb) {
+  return cb(arr.length)
   // getLength passes the length of the array into the callback.
+  
 }
+getLength (items, cb)
```
Velsu, line 31: You would want to replace `cb` with some function that You want to call as callback, for example `console.log`.

Velsu (commented):
Great job on the assignments Jake. You have completed the array, callbacks and closures. Your code is clean and this is fine job 👍

PR #330, opened 2018-09-11
Sean Flannigan - JavaScript-II

sean-one:


jeffyjkang:
great work on the callback and array-methods

PR #332, opened 2018-09-11
Adam Hinckley

adamhinckley:


assignments/array-methods.js
```diff
@@ -2,82 +2,517 @@
...
+})
+
 console.log(fullName);
 
 // ==== Challenge 2: Use .map() ====
 // The event director needs to have all the runner's first names converted to uppercase because the director BECAME DRUNK WITH POWER. Convert each first name into all caps and log the result
 let allCaps = [];
-console.log(allCaps); 
+runners.map(function (runners) {
+  allCaps.push(runners.first_name.toUpperCase());
```
Velsu, line 470: No need to create empty array beforehand and using `push` to insert data into it. You are using `.map` which creates new array with modified data already so just use `allCaps = runners.map(...)`

assignments/array-methods.js
```diff
@@ -2,82 +2,517 @@
...
+})
+console.log(allCaps);
 
 // ==== Challenge 3: Use .filter() ====
 // The large shirts won't be available for the event due to an ordering issue.  Get a list of runners with large sized shirts so they can choose a different size. Return an array named largeShirts that contains information about the runners that have a shirt size of L and log the result
 let largeShirts = [];
+runners.filter(function (runners) {
+  if (runners.shirt_size === 'L') {
+    largeShirts.push(runners.first_name + ' ' + runners.last_name)
```
Velsu, line 479: Same as allCaps, no need to push, just create a new variable and run filter on it.

assignments/array-methods.js
```diff
@@ -2,82 +2,517 @@
...
+  }
+})
 console.log(largeShirts);
 
 // ==== Challenge 4: Use .reduce() ====
 // The donations need to be tallied up and reported for tax purposes. Add up all the donations into a ticketPriceTotal array and log the result
 let ticketPriceTotal = [];
+runners.reduce(function (total, runners) {
+  ticketPriceTotal.push(runners.donation);
```
Velsu, line 488: Aaaand again, similar to map and filter, no need to push and declare empty array.

assignments/closure.js
```diff
@@ -1,11 +1,32 @@
...
+  }
+  diet()
+}
+console.log(favoriteAnimal())
 // ==== Challenge 2: Create a counter function ====
 const counter = () => {
-  // Return a function that when invoked increments and returns a counter variable.
+  let count = 0;
+
+  function counting() {
```
Velsu, line 19: Counter is not working properly, it seems it returns undefined. Maybe make an arrow function that returns self increment?

Velsu (commented):
Great job on this one Adam. I noted some improvements to the code, especially with map and reduce (no need to make new array manually with these) and an issue with counter.

Except that everything is great and good job 👍

PR #342, opened 2018-09-11
Hunter Smith - JavaScript-II Project

Hunter315:
@kamry-bowman

### DIFF
--- a/assignments/array-methods.js
+++ b/assignments/array-methods.js
@@ -62,23 +62,61 @@ console.log(fullName);
 // ==== Challenge 2: Use .map() ====
 // The event director needs to have all the runner's first names converted to uppercase because the director BECAME DRUNK WITH POWER. Convert each first name into all caps and log the result
 let allCaps = [];
+allCaps = runners.map(runner => { return `${runner.first_name.toUpperCase()}`})
 console.log(allCaps); 
 
 // ==== Challenge 3: Use .filter() ====
 // The large shirts won't be available for the event due to an ordering issue.  Get a list of runners with large sized shirts so they can choose a different size. Return an array named largeShirts that contains information about the runners that have a shirt size of L and log the result
 let largeShirts = [];
-console.log(largeShirts);
+largeShirts = runners.filter(runner => runner.shirt_size === "L");
+console.log(JSON.stringify(largeShirts));
+// console.table(largeShirts);
+// alert(largeShirts);
 
 // ==== Challenge 4: Use .reduce() ====
 // The donations need to be tallied up and reported for tax purposes. Add up all the donations into a ticketPriceTotal array and log the result
 let ticketPriceTotal = [];
+ticketPriceTotal = runners.reduce((ticketSales, runner) => {return ticketSales += runner.donation}, 0);
 console.log(ticketPriceTotal);
 
 // ==== Challenge 5: Be Creative ====
 // Now that you have used .forEach(), .map(), .filter(), and .reduce().  I want you to think of potential problems you could solve given the data set and the 5k fun run theme.  Try to create and then solve 3 unique problems using one or many of the array methods listed above.
 
 // Problem 1
+// Make an email list for the marketing team
+let emailList = function(runners){
+    return runners.map(runner => runner.email);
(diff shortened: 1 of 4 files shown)


assignments/array-methods.js
```diff
@@ -56,28 +56,67 @@ const runners = [{"id":1,"first_name":"Charmain","last_name":"Seiler","email":"c
 // ==== Challenge 1: Use .forEach() ====
 // The event director needs both the first and last names of each runner for their running bibs.  Combine both the first and last names into a new array called fullName. 
 let fullName = [];
+runners.forEach(runner => {fullName.push(`${runner.first_name} ${runner.last_name}`)});
 console.log(fullName);
 
 // ==== Challenge 2: Use .map() ====
 // The event director needs to have all the runner's first names converted to uppercase because the director BECAME DRUNK WITH POWER. Convert each first name into all caps and log the result
 let allCaps = [];
+allCaps = runners.map(runner => { return `${runner.first_name.toUpperCase()}`})
```
kamry-bowman, line 65: if you're doing this in a single line, I would recommend like this:
```
allCaps = runners.map(runner => `${runner.first_name.toUpperCase()}`)
```

If you use the curly brackets, I would recommend breaking it up like this: 
```
allCaps = runners.map(runner => {
    return `${runner.first_name.toUpperCase()}`;
})
```

assignments/array-methods.js
```diff
@@ -56,28 +56,67 @@ const runners = [{"id":1,"first_name":"Charmain","last_name":"Seiler","email":"c
...
+// Make an email list for the marketing team
+let emailList = function(runners){
+    return runners.map(runner => runner.email);
+    
+}
+console.log(emailList(runners).sort());
 
 // Problem 2
-
+// add up the different number of shirt sizes worn.
```
kamry-bowman, line 94: nice idea on this one. Remember you were supposed to use an array method though! Also, here's a different way to approach it using an object instead of an array to keep track of the count;
what if you tried this? 
```
let shirtSizes = {};
for (i =0; i < runners.length; i++){
    const size = runners[i].shirt_size;
    if (shirtSizes[size] === undefined) {
        shirtSizes[size] = 1;
    } else {
        shirtSizes[size] += 1;
    }
 
};
```

assignments/callbacks.js
```diff
@@ -20,33 +20,58 @@ const items = ['Pencil', 'Notebook', 'yo-yo', 'Gum'];
...
+  return cb(arr.length);
 }
+getLength(items, function(length){
+  console.log(length)
+})
+
 
 function last(arr, cb) {
   // last passes the last item of the array into the callback.
+  return cb(arr.pop());
```
kamry-bowman, line 36: this one is not a good choice, because pop mutates your original array, so it will remove the last item from your array each time you call it, which it's not supposed to do. Try `arr[arr.length - 1]` instead

assignments/callbacks.js
```diff
@@ -20,33 +20,58 @@ const items = ['Pencil', 'Notebook', 'yo-yo', 'Gum'];
...
+  return cb(x*y);
 }
+multiplyNums (3, 5, function(mult){
+  console.log(mult);
+});
 
 function contains(item, list, cb) {
   // contains checks if an item is present inside of the given array/list.
   // Pass true to the callback if it is, otherwise pass false.
+  return cb(list.includes(item));
```
kamry-bowman, line 63: nice one-liner

assignments/closure.js
```diff
@@ -1,14 +1,35 @@
 // ==== Challenge 1: Write your own closure ====
 // Write a simple closure of your own creation.  Keep it simple!
+const niceHello = () => {
+    return "Hello, Hunter! You are programming so well today!"
+}
+console.log(niceHello());
```
kamry-bowman, line 6: I don't think this uses closure. Remember, closures require a function to be nested inside another. DM me if you'd like to discuss it further

kamry-bowman (commented):
Great job on this one! A few notes, let me know if you'd like to discuss at all! Also look into arrow functions with implicit returns

PR #347, opened 2018-09-11
Hargobind Atwal

ethiapath:


CookieMonsta89:
Great work Hargobind! Your code was very clean and easy to read.  I appreciate that.  Keep up the good work.  Everything passed its tests so great job.  You can definitely tell you have worked with JS before.

PR #351, opened 2018-09-11
Melinda Yeager

Mindysueasaurus:


CookieMonsta89:
Great job Mindy on your project.  I noticed that on callbacks.js, line 29, you were missing a closing parenthesis that was causing an error.  I added it and it worked just fine.  Keep up the good work.

PR #354, opened 2018-09-11
Katie Fitzpatrick - JavaScript-II

KatieFizzy:


assignments/array-methods.js
```diff
@@ -54,30 +54,74 @@ const runners = [{"id":1,"first_name":"Charmain","last_name":"Seiler","email":"c
...
 // ==== Challenge 2: Use .map() ====
-// The event director needs to have all the runner's first names converted to uppercase because the director BECAME DRUNK WITH POWER. Convert each first name into all caps and log the result
+// The event director needs to have all the runner's first names converted to uppercase because the
+// director BECAME DRUNK WITH POWER. Convert each first name into all caps and log the result
 let allCaps = [];
+runners.map(function(item) {
+    allCaps.push(item.first_name.toUpperCase())
+})
+
 console.log(allCaps);
```
kamry-bowman, line 77: this is more how you would use forEach, not map. Here's the solution code for this one: 
```
let allCaps = runners.map((person) => {
  return person.first_name.toUpperCase();
});
console.log(allCaps); 
```

assignments/array-methods.js
```diff
@@ -54,30 +54,74 @@ const runners = [{"id":1,"first_name":"Charmain","last_name":"Seiler","email":"c
...
 
-// Problem 3
\ No newline at end of file
+// Problem 3 Skinix is a big corporate donor and the 
+//5K wants to have a special cocktail hour for their employees. Make a contact lists.
+let skinixEmployees= runners.filter((employees) => {
+	return employees.company_name == "Skinix";
+})
+console.log(skinixEmployees); 
+//apparently this will be an awkard 3 person cocktail hour
```
kamry-bowman, line 127: haha a reasonable risk of just one person showing up, that's always fun

assignments/callbacks.js
```diff
@@ -20,28 +21,57 @@ const items = ['Pencil', 'Notebook', 'yo-yo', 'Gum'];
   });
 
 */
+const items = ['Pencil', 'Notebook', 'yo-yo', 'Gum'];
+
 
 
 function getLength(arr, cb) {
+  return cb(arr.length+1);
```
kamry-bowman, line 29: I think the general style preferred is to have spaces between operators `arr.length + 1`

assignments/callbacks.js
```diff
@@ -20,28 +21,57 @@ const items = ['Pencil', 'Notebook', 'yo-yo', 'Gum'];
...
 function multiplyNums(x, y, cb) {
+  return cb(x*y);
   // multiplyNums multiplies two numbers and passes the result to the callback.
 }
 
+multiplyNums(5,6,function(multiply) {
+  console.log(multiply);
+})
+
 function contains(item, list, cb) {
```
kamry-bowman, line 61: great job on this one. The only thing I would add is that it should also return cb(false) after the loop. That way, the function still gets called with false if the items not in the list, which is what the specification calls for.

assignments/closure.js
```diff
@@ -1,10 +1,40 @@
...
+        return 'who knows what kind of animal that is!'
+      }
+    }
+    return animal;
+  };
+  
+let animalType = animalNoise();
+
+console.log(animalType('quack'));
+console.log(animalType('meow'));
```
kamry-bowman, line 19: I really like this example! Note though, that because `soundMade` is the parameter for the `animal` function, it actually shadows the `soundMade` variable in the animalNoise closure, so animal never makes use of the closure! DM me to discuss more if you would like, as it's a useful thing to understand.

kamry-bowman (commented):
Awesome job on this one, great style and understanding of concepts. Let me know if you have questions on any comments!

PR #358, opened 2018-09-11
JavaScript-II By - Michael Checo

MichaelCheco:


KevinBrack:
array-methods.js was excellent. Callbacks.js is missing the contains function. closure.js was great. Really nice work.

PR #371, opened 2018-09-11
Blake Koennecker - JavaScript-II

koennecker:


Velsu (commented):
Solid job on the assingments Blake. Clean code, all challenges solved successfully and as usualy great git workflow 👍

PR #382, opened 2018-09-11
Oleksandr Stasyev

CSDisciple:
Finished arrays and callbacks


CookieMonsta89:
Great job Olek! Mostly everything works.  I would like for you to take a look at your callbacks because I noticed you aren't actually using a cb function.  For example, how would you test your multiplyNums function?  There isn't a cb function to pass through.  I wrote an example of what it might look like if you were using a callback.  Compare it to yours.  Keep up the good work.    ```let multi = (x, y) => {
    return x * y;
}
    
    return cb(x,y);
}
console.log(multiplyNums(2,2,multi))```

PR #388, opened 2018-09-11
Javascript II - Edd Burke

bummings:


KevinBrack:
closure.js is missing the second challenge. Other than that doing good. Keep it up

PR #389, opened 2018-09-11
JS 2

Bryan-GF:
@Jeffyjkang


jeffyjkang:
great work on the array methods and callbacks, i think you may have to revisit/review closures

PR #391, opened 2018-09-11
Angelo Deleon - JAVASCRIPT-II

RaDeleon:
@kamry_bowman

### DIFF
--- a/assignments/array-methods.js
+++ b/assignments/array-methods.js
@@ -94,11 +94,20 @@ console.log(largeShirts);
 // The donations need to be tallied up and reported for tax purposes. Add up all the donations into a ticketPriceTotal array and log the result
 let ticketPriceTotal = [];
 
-runners.reduce((previous, current) => {
-    
-}, initial)
+function donations(arr){
+	let i = 0;
+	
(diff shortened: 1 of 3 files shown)


assignments/array-methods.js
```diff
@@ -56,28 +56,77 @@ const runners = [{"id":1,"first_name":"Charmain","last_name":"Seiler","email":"c
...
 
+
+
+
 // ==== Challenge 2: Use .map() ====
 // The event director needs to have all the runner's first names converted to uppercase because the director BECAME DRUNK WITH POWER. Convert each first name into all caps and log the result
 let allCaps = [];
+
+runners.map((item) => {
+    allCaps.push(item.first_name.toUpperCase());
```
Velsu, line 73: No need to use `push` when using `.map()`, as it already creates a copy of new array with modified data from the old one, so better aproach would be something like this:

```
allCaps = runners.map(item => {
  return item.first_name.toUpperCase();
});
```

Velsu (commented):
Good job on the assignment Angelo. All exercises done properly. You should add some logs to callbacks.js to see if methods are actually working but i tested them and they return proper results.

I also added a comment regading .map() method, because You used it as regular forEach loop which is not necessary.

Overall great job 👍

PR #398, opened 2018-09-11
Quan Nguyen_ JavaScriptII

qtn2932:
@sumi419


sumi419:
Great job Quan! I'm proud of you for finishing this too 🎉

PR #403, opened 2018-09-11
not complete

Thacker-E:
@ta1grr

### DIFF
--- a/assignments/array-methods.js
+++ b/assignments/array-methods.js
@@ -55,12 +55,15 @@ const runners = [{"id":1,"first_name":"Charmain","last_name":"Seiler","email":"c
 
 // ==== Challenge 1: Use .forEach() ====
 // The event director needs both the first and last names of each runner for their running bibs.  Combine both the first and last names into a new array called fullName. 
-let fullName = [];
+fullName.foreach(function(fullName)){
+   
+};
 console.log(fullName);
+
 
 // ==== Challenge 2: Use .map() ====
 // The event director needs to have all the runner's first names converted to uppercase because the director BECAME DRUNK WITH POWER. Convert each first name into all caps and log the result
-let allCaps = [];
+let allCaps = runners.map(     =>first_name.toUpperCase());
 console.log(allCaps); 
 
 // ==== Challenge 3: Use .filter() ====
--- a/assignments/callbacks.js
+++ b/assignments/callbacks.js
@@ -23,8 +23,12 @@ const items = ['Pencil', 'Notebook', 'yo-yo', 'Gum'];
 
 
 function getLength(arr, cb) {
+  return cb(arr[1]);
   // getLength passes the length of the array into the callback.
 }
+getLength(items, function('second'))
+console.log('second')
+});
 
 function last(arr, cb) {
   // last passes the last item of the array into the callback.


assignments/array-methods.js
```diff
@@ -55,12 +55,15 @@ const runners = [{"id":1,"first_name":"Charmain","last_name":"Seiler","email":"c
 
 // ==== Challenge 1: Use .forEach() ====
 // The event director needs both the first and last names of each runner for their running bibs.  Combine both the first and last names into a new array called fullName. 
-let fullName = [];
+fullName.foreach(function(fullName)){
+
```
Ta1grr, line 59: You were super close, you just needed to put in the instructions for the `.foreach`
```
let fullName = [] //assigning an empty array to the variable named fullName.
//We're applying .foreach() to runners and inside we're creating a function to do a job to each object.
let getFirstLastName = runners.forEach(function(index) { 
//We're assign .push to fullName and .foreach will run this command to each object and push it into the variable fullName.
fullName.push(index.first_name + ' ' + index.last_name); 
});

console.log(fullName); //console.log to get the array of fullName with first_name and last_name populate inside of it.
```

assignments/array-methods.js
```diff
@@ -55,12 +55,15 @@ const runners = [{"id":1,"first_name":"Charmain","last_name":"Seiler","email":"c
...
+fullName.foreach(function(fullName)){
+   
+};
 console.log(fullName);
 
+
 // ==== Challenge 2: Use .map() ====
 // The event director needs to have all the runner's first names converted to uppercase because the director BECAME DRUNK WITH POWER. Convert each first name into all caps and log the result
-let allCaps = [];
+let allCaps = runners.map(     =>first_name.toUpperCase());
```
Ta1grr, line 66: not sure what you were doing here, but here's an example with regular function.
Example:
```
let allCaps = runners.map(function(index) { //map will take everything and put into a new array, you can put anything for the parameter.
  return index.first_name.toUpperCase(); //here, we're calling the parameter from the parenthesis and calling the first_name from each object and uppercase them before putting it into the array.
});

console.log(allCaps); //should show all first names capitalized
```

assignments/callbacks.js
```diff
@@ -23,8 +23,12 @@ const items = ['Pencil', 'Notebook', 'yo-yo', 'Gum'];
 
 
 function getLength(arr, cb) {
+  return cb(arr[1]);
```
Ta1grr, line 26: You're hardcoding an array index, so it'll always refer to that location. Please refer to this documentation to fully grasp `.length` https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/length
Solution:
```
function getLength(arr, cb) {
  // getLength passes the length of the array into the callback. Imagine the callback(cb) function as Gru from despicable me.
  console.log(cb(arr));
}

let arrayLength = function(items) { //this is your minion
  return items.length;
}

getLength(items,arrayLength);
```
Please read this article to give you another view of callbacks.
https://medium.freecodecamp.com/javascript-callbacks-explained-using-minions-da272f4d9bcd

Ta1grr (commented):
### Great
 - It's normal to struggle, otherwise you won't have questions.
## Requested Improvements
 - Check the comments for improvements.
 - Rewatch the lecture and make another attempt on the project
## Questions
 - None at the moment.
## Rating: {1-3}
  - 1 (Does not understand, does not meet expectations)

PR #410, opened 2018-09-17
barbara hunter

cmdncode:


Velsu (commented):
Barbara, good job on the closures exercise, but it seems You did not do all the other assignments.

Missing exercises:
- Callbacks
- Array methods

I strongly recommend to revisit these and try to complete them.

PR #415, opened 2018-10-16
Grace Kang

bananaboots:


assignments/callbacks.js
```diff
@@ -23,30 +25,28 @@ const items = ['Pencil', 'Notebook', 'yo-yo', 'Gum'];
...
 }
 
 /* STRETCH PROBLEM */
 
 function removeDuplicates(array, cb) {
-  // removeDuplicates removes all duplicate values from the given array.
-  // Pass the duplicate free array to the callback function.
-  // Do not mutate the original array.
+  let unique = [...new Set(array)];
+  return cb(unique);
```
johnoro, line 51: You could shorten this by foregoing the variable:
```javascript
return cb([...new Set(array)]);
```
although there is an argument to be made for clarity by having the extra variable (given that you name it well, which you did).

assignments/array-methods.js
```diff
@@ -56,28 +56,45 @@ const runners = [{"id":1,"first_name":"Charmain","last_name":"Seiler","email":"c
 // ==== Challenge 1: Use .forEach() ====
 // The event director needs both the first and last names of each runner for their running bibs.  Combine both the first and last names into a new array called fullName. 
 let fullName = [];
+runners.forEach((cur) => {
+    fullName.push(`${cur["first_name"]} ${cur["last_name"]}`);
```
johnoro, line 60: You could access the `cur` properties with dot notation (which would be a little shorter to type here):
```javascript
fullName.push(`${cur.first_name} ${cur.last_name}`);
```

assignments/array-methods.js
```diff
@@ -56,28 +56,45 @@ const runners = [{"id":1,"first_name":"Charmain","last_name":"Seiler","email":"c
...
+});
```
johnoro, line 61: You could also use [object destructuring](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment#Object_destructuring) to get ahold of the values in this way:
```javascript
runners.forEach(({ first_name, last_name }) => {
    fullName.push(`${first_name} ${last_name}`);
});
```

assignments/array-methods.js
```diff
@@ -56,28 +56,45 @@ const runners = [{"id":1,"first_name":"Charmain","last_name":"Seiler","email":"c
...
 // ==== Challenge 3: Use .filter() ====
 // The large shirts won't be available for the event due to an ordering issue.  Get a list of runners with large sized shirts so they can choose a different size. Return an array named largeShirts that contains information about the runners that have a shirt size of L and log the result
 let largeShirts = [];
+largeShirts = runners.filter((cur) => (cur["shirt_size"] === "L"));
 console.log(largeShirts);
 
 // ==== Challenge 4: Use .reduce() ====
 // The donations need to be tallied up and reported for tax purposes. Add up all the donations into a ticketPriceTotal array and log the result
 let ticketPriceTotal = [];
+ticketPriceTotal.push(runners.reduce((acc, cur) => acc + parseInt(cur["donation"]), 0));
```
johnoro, line 79: Why use `push` here? You could just reassign `ticketTotalPrice` to be the sum returned by your `reduce`.

johnoro, line 79: Also, why use `parseInt`? All of our donations are numbers already!

assignments/array-methods.js
```diff
@@ -56,28 +56,45 @@ const runners = [{"id":1,"first_name":"Charmain","last_name":"Seiler","email":"c
...
+let lastNames = [];
+runners.forEach((cur) => lastNames.push(`${cur["last_name"]} ${cur["first_name"]}`));
+lastNames.sort();
+console.log(lastNames);
 
-// Problem 3
\ No newline at end of file
+// Problem 3
+// We decidedly only care about runners with donations of over $100. Let's put them on the REAL list of runners!
+let realRunners = runners.filter((cur) => (parseInt(cur["donation"]) > 100));
```
johnoro, line 99: Same question about `parseInt` here!

assignments/closure.js
```diff
@@ -14,7 +25,21 @@ const counter = () => {
...
 // ==== Challenge 3: Create a counter function with an object that can increment and decrement ====
 const counterFactory = () => {
+  let obj = {"count": 0};
+  obj["increment"] = function(){
+    return ++this.count;
+  }
+  obj["decrement"] = function(){
+    return --this.count;
+  }
+  return obj;
```
johnoro, line 35: This gets the job done, but is a little lengthy and isn't taking full advantage of closures. You could've done it this way as well:
```javascript
const counterFactory = () => {
  let count = 0;
  return {
    increment: () => ++count;
    decrement: () => --count;
  };
}
```

johnoro (commented):
Awesome work! I love all the arrow functions.

PR #418, opened 2018-10-16
Nedim Omerovic - JavaScript II

Civoremo:


assignments/array-methods.js
```diff
@@ -56,28 +57,65 @@ const runners = [{"id":1,"first_name":"Charmain","last_name":"Seiler","email":"c
...
+        largeShirts.push(arr);
+    }
+});
 console.log(largeShirts);
 
 // ==== Challenge 4: Use .reduce() ====
 // The donations need to be tallied up and reported for tax purposes. Add up all the donations into a ticketPriceTotal array and log the result
 let ticketPriceTotal = [];
+const totalDonation = 
+runners.reduce(function(money, arr){
```
trevorfehrman, line 88: Try some more semantic parameters.  The second one, for example is the element of the array, not the array.  Something like "runner" or "racer" might be reasonable.

assignments/closure.js
```diff
@@ -1,14 +1,56 @@
...
 // Write a simple closure of your own creation.  Keep it simple!
+function myFunction(param){
+  const month = param;
+  console.log(`The orangest month of the year is ${param}`);
 
+  function day(){
+    const favDay = "31st";
+    console.log(`Halloween is on ${month}, ${favDay}.`);
+
+    function today() {
```
trevorfehrman, line 11: This is great

trevorfehrman (commented):
Not much to suggest here, Great job sir.

PR #422, opened 2018-10-16
Eileen Cuevas

eileencuevas:


assignments/callbacks.js
```diff
@@ -19,34 +20,79 @@ const items = ['Pencil', 'Notebook', 'yo-yo', 'Gum'];
...
 
 function removeDuplicates(array, cb) {
   // removeDuplicates removes all duplicate values from the given array.
   // Pass the duplicate free array to the callback function.
   // Do not mutate the original array.
+
+  let noDupes = [];
+  
+  noDupes = array.filter(function(item, index) {
+      return index <= array.indexOf(item);
```
trevorfehrman, line 88: So in case you didn't already know the reason this works is because indexOf returns -1 if the argument isn't in the array upon which the method is invoked.  .includes() is probably the best tool for this, or Set.

Great job though.

assignments/closure.js
```diff
@@ -1,14 +1,32 @@
 // ==== Challenge 1: Write your own closure ====
 // Write a simple closure of your own creation.  Keep it simple!
 
+const currentCash = 0;
+
+function currentMoney(name) {
+  const sentence = 'your wallet has this much in it:'
+
+  function createStatement() {
```
trevorfehrman, line 9: Perfect

assignments/closure.js
```diff
@@ -1,14 +1,32 @@
...
+
+currentMoney('me');
 
 // ==== Challenge 2: Create a counter function ====
-const counter = () => {
+const counter = () => { 
   // Return a function that when invoked increments and returns a counter variable.
+  let count = 0;
+  
+  return () => (++count);
```
trevorfehrman, line 23: can you explain why this works?  ARE U SUUUUUREE?!?!?!?!?!?!?!?!!?

mwhahaha

PR #424, opened 2018-10-16
Josh Zieger

ghost:


assignments/array-methods.js
```diff
@@ -56,28 +56,65 @@ const runners = [{"id":1,"first_name":"Charmain","last_name":"Seiler","email":"c
...
 console.log(allCaps); 
 
 // ==== Challenge 3: Use .filter() ====
 // The large shirts won't be available for the event due to an ordering issue.  Get a list of runners with large sized shirts so they can choose a different size. Return an array named largeShirts that contains information about the runners that have a shirt size of L and log the result
 let largeShirts = [];
+function isLarge(runners){
+    if (runners.shirt_size === 'L'){
+        largeShirts.push(`${runners.first_name} ${runners.last_name},  ${runners.email}, ${runners.company_name} `) }
+    }
+const runnersWithLargeShirts = runners.filter(isLarge);
```
Kamalnrf, line 79: you can also do it this way `runners.filter(runner => runner.shirt_size === 'L')`

assignments/array-methods.js
```diff
@@ -56,28 +56,65 @@ const runners = [{"id":1,"first_name":"Charmain","last_name":"Seiler","email":"c
...
 
-// Problem 3
\ No newline at end of file
+// Problem 3
+// List all companies involved in alphabetical order for promotional material
+// I know this is similar to problem #2, but I was having a hard time coming up with anything and running out of time.
+let companyList = [];
+runners.forEach(function(runner) {
+    companyList.push(`${runner.company_name}: $${runner.donation}`);
+    companyList.sort();
```
Kamalnrf, line 118: you are sorting in every iteration this way you are doing a lot more operations instead you can sort everything after the adding them all (outside for each)

```
companyList = runners.forEach(function(runner) {
    companyList.push(`${runner.company_name}: $${runner.donation}`);
})
companyList = companyList.sort()

File: assignments/callbacks.js [hunk 3]
```diff
@@ -21,26 +21,38 @@ const items = ['Pencil', 'Notebook', 'yo-yo', 'Gum'];
...
 function multiplyNums(x, y, cb) {
   // multiplyNums multiplies two numbers and passes the result to the callback.
+  return cb(x * y);
 }
 
 function contains(item, list, cb) {
   // contains checks if an item is present inside of the given array/list.
   // Pass true to the callback if it is, otherwise pass false.
+  cb(item, function(items){
+    for (let i=0; i<list.length; i++){
```
Kamalnrf, line 48: I appreciate it when you are using loops instead of inbuilt functions, it shows that you are getting comfortable in applying the logic.

Just to show other way you can use `includes` array function, remember that underneath it has the same logic you wrote.
```
  if (!list.includes(item))
    return cb(false)
   return true
}

PR #425, opened 2018-10-16
Jonathan Heinz

jheinz1999:


assignments/callbacks.js
```diff
@@ -24,29 +24,50 @@ const items = ['Pencil', 'Notebook', 'yo-yo', 'Gum'];
...
+multiplyNums(10, 20, product => console.log(`The product is ${product}`));
+
 function contains(item, list, cb) {
   // contains checks if an item is present inside of the given array/list.
   // Pass true to the callback if it is, otherwise pass false.
+  cb(list.includes(item));
 }
 
+contains('Pencil', items, contained => contained ? console.log("Item is in array") : console.log("Item is not in array"));
+contains('Pen', items, contained => contained ? console.log("Item is in array") : console.log("Item is not in array"));
```
johnoro, line 60: You could make a named function for the callback that you use twice to reduce repetition; i.e.:
```javascript
const logItemPresence = contained => contained 
  ? console.log("Item is in array") 
  : console.log("Item is not in array");
contains('Pencil', items, contained => logItemPresence);
contains('Pen', items, logItemPresence);
```

johnoro: You could also shorten it by using `console.log` only once:
```javascript
const logItemPresence = contained => console.log(
  contained 
    ? "Item is in array" 
    : "Item is not in array"
);
contains('Pencil', items, contained => logItemPresence);
contains('Pen', items, logItemPresence);
```

assignments/array-methods.js
```diff
@@ -54,30 +54,45 @@ const runners = [{"id":1,"first_name":"Charmain","last_name":"Seiler","email":"c
 {"id":50,"first_name":"Shell","last_name":"Baine","email":"sbaine1d@intel.com","shirt_size":"M","company_name":"Gabtype","donation":171}];
 
 // ==== Challenge 1: Use .forEach() ====
-// The event director needs both the first and last names of each runner for their running bibs.  Combine both the first and last names into a new array called fullName. 
+// The event director needs both the first and last names of each runner for their running bibs.  Combine both the first and last names into a new array called fullName.
 let fullName = [];
+runners.forEach(item => fullName.push(`${item.first_name} ${item.last_name}`));
```
johnoro, line 59: You could use [object destructuring](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment#Object_destructuring) here:
```javascript
 runners.forEach(({ first_name, last_name }) => fullName.push(`${first_name} ${last_name}`)); 
```

assignments/array-methods.js
```diff
@@ -54,30 +54,45 @@ const runners = [{"id":1,"first_name":"Charmain","last_name":"Seiler","email":"c
...
+console.log(cheapskates);
 
-// Problem 3
\ No newline at end of file
+// Problem 3: How many companies are at the run?
+let companies = [];
+runners.forEach(item => {
+  if (!companies.includes(item.company_name))
+    companies.push(item.company_name)
+});
```
johnoro, line 97: Here is a good example of where object destructuring is nice for brevity since you're accessing the same key multiple times:
```javascript
runners.forEach(({ company_name }) => {
  if (!companies.includes(company_name))
    companies.push(company_name)
});
```

assignments/closure.js
```diff
@@ -5,16 +5,48 @@
...
+    },
+    decrement: function() {
+      console.log(--counter);
+    }
+  }
 };
+
+console.log("\n");
+
+const newerCounter = counterFactory();
```
johnoro, line 40: `newerCounter` is a very good variable name; it's definitely among the greatest of the greats.

assignments/closure.js
```diff
@@ -5,16 +5,48 @@
...
   // `increment` should increment a counter variable in closure scope and return it.
   // `decrement` should decrement the counter variable and return it.
+  let counter = 0;
+  return {
+    increment: function() {
+      console.log(++counter);
+    },
+    decrement: function() {
+      console.log(--counter);
+    }
```
johnoro, line 34: You could shorten these with arrow functions; for example:
```javascript
decrement: () => {
  console.log(--counter);
}
```
or even

```javascript
decrement: () => console.log(--counter);
```
although the last could be considered a no-no since you're not necessarily using that value that you're _implicitly_ returning (but it's fine in this case, and JS will let you do it too)

johnoro (commented):
Awesome work!

PR #426, opened 2018-10-16
First Submission - bare minimum tasks completed

gittc100:


assignments/callbacks.js
```diff
@@ -23,30 +23,67 @@ const items = ['Pencil', 'Notebook', 'yo-yo', 'Gum'];
 
 
 function getLength(arr, cb) {
-  // getLength passes the length of the array into the callback.
+  return cb(arr.length);
 }
 
+getLength(items, function(x){
+  console.log(x);
+});
```
johnoro, line 31: You reuse this callback throughout; it'd be easier to assign it to a variable and also reduce repetition:
```javascript
// ...
const log = function(x){
  console.log(x);
}

getLength(items, log);
last(items, log);
// ...
```

johnoro: The other thing to note here is that `log` **is** a function. Specifically, it's a method on the `console` object. You could instead pass it as your callback function since it's unnecessary to pass it via another function:
```javascript
// ...
last(items, console.log);
// ...
```

assignments/array-methods.js
```diff
@@ -56,27 +56,47 @@ const runners = [{"id":1,"first_name":"Charmain","last_name":"Seiler","email":"c
...
-// Problem 1
+// Problem 1 - Create an Array Containing runner name and email in object form
+
+let arrObj1 = [];
+arrObj1.push(runners.map((item) => {
+    return {"first_name": item.first_name,"last_name":item.last_name,"email":item.email};
+}));
+console.log(arrObj1);
 
 // Problem 2
```
johnoro, line 101: Create a couple more problems and solve them! I'm assuming I just caught you when you weren't quite done here.

johnoro (commented):
Awesome work!

PR #427, opened 2018-10-16
Nick Stricker

NickStrick:


assignments/callbacks.js
```diff
@@ -24,29 +24,71 @@ const items = ['Pencil', 'Notebook', 'yo-yo', 'Gum'];
...
+})
+
+  return cb(uniqueArray);
+}
+
+function consoleL(thing){
+  console.log(thing);
 }
+
+removeDuplicates([1,1,5,3,3,4,4,5], consoleL);
```
johnoro, line 93: You could forego the new function here and simply use `console.log` as your callback function:
```javascript
 removeDuplicates([1,1,5,3,3,4,4,5], console.log); 
```
`log` is just method on the `console` object, so you can plug it in as a callback just like any other function!

assignments/stretch-function-conversion.js
```diff
@@ -1,23 +1,29 @@
 // Take the commented ES5 syntax and convert it to ES6 arrow Syntax
 
 // let myFunction = function () {};
+let myFunction = () => {};
 
 // let anotherFunction = function (param) {
 //   return param;
 // };
+let anotherFunction = (param) => param
```
johnoro, line 9: You can drop the parentheses here since you only have one parameter and aren't doing anything fancy with it:
```javascript
 let anotherFunction = param => param ;
```

johnoro (commented):
Great work!

PR #431, opened 2018-10-16
Austin Blake

dyiar:
Finished Callbacks

### DIFF
--- a/assignments/array-methods.js
+++ b/assignments/array-methods.js
@@ -1,4 +1,6 @@
-// A local community center is holding a fund rasising 5k fun run and has invited 50 small businesses to make a small donation on their behalf for some much needed updates to their facilities.  Each business has assigned a representative to attend the event along with a small donation.
+// A local community center is holding a fund rasising 5k fun run and has invited 50 small businesses to make
+// a small donation on their behalf for some much needed updates to their facilities.  Each business has assigned 
+// a representative to attend the event along with a small donation.
 
 // Scroll to the bottom of the list to use some advanced array methods to help the event director gather some information from the businesses.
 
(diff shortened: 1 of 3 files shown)


assignments/array-methods.js
```diff
@@ -54,30 +56,71 @@ const runners = [{"id":1,"first_name":"Charmain","last_name":"Seiler","email":"c
...
+
 console.log(allCaps); 
 
 // ==== Challenge 3: Use .filter() ====
-// The large shirts won't be available for the event due to an ordering issue.  Get a list of runners with large sized shirts so they can choose a different size. Return an array named largeShirts that contains information about the runners that have a shirt size of L and log the result
-let largeShirts = [];
+// The large shirts won't be available for the event due to an ordering issue. 
+// Get a list of runners with large sized shirts so they can choose a different size. 
+// Return an array named largeShirts that contains information about the runners that have a shirt size of L and log the result
+let largeShirts = runners.filter((size) => {
```
trevorfehrman, line 82: Nitpicky but I'd name my parameter shirt here since that's what you're actually iterating over.  Then your return statement would be `shirt.shirt_size`

assignments/array-methods.js
```diff
@@ -54,30 +56,71 @@ const runners = [{"id":1,"first_name":"Charmain","last_name":"Seiler","email":"c
...
+    return size.shirt_size === 'L'
+});
+
 console.log(largeShirts);
 
 // ==== Challenge 4: Use .reduce() ====
 // The donations need to be tallied up and reported for tax purposes. Add up all the donations into a ticketPriceTotal array and log the result
-let ticketPriceTotal = [];
+let ticketPriceTotal = runners.reduce(function(totalPrice, price) {
```
trevorfehrman, line 90: Again a parameter that accounts for the whole object might be preferable here.

assignments/callbacks.js
```diff
@@ -1,52 +1,116 @@
...
+const animals = ["cat", "cat", "dog", "giraffe", "cat", "dog", "pig", "cow"];
+
 function removeDuplicates(array, cb) {
   // removeDuplicates removes all duplicate values from the given array.
   // Pass the duplicate free array to the callback function.
   // Do not mutate the original array.
+  let noDupes = [];
+  
+  noDupes = array.filter(function(item, index) {
+    return array.indexOf(item) >= index;
```
trevorfehrman, line 88: Okay I now know why this works.  .includes() is better, this is still valid though.

indexOf returns -1 if the argument isn't in the array.

assignments/closure.js
```diff
@@ -1,11 +1,57 @@
...
+// add();
+// add();
+// add();
+
+// console.log(counter);
+
 const counter = () => {
+  let count = 0;
+  
+  return () => (++count);
```
trevorfehrman, line 37: I think we discussed this at length yesterday so I'll spare you

PR #432, opened 2018-10-16
Daniel Mendez Javascript-II

dan2014:
working on closures

### DIFF
--- a/assignments/array-methods.js
+++ b/assignments/array-methods.js
@@ -56,28 +56,52 @@ const runners = [{"id":1,"first_name":"Charmain","last_name":"Seiler","email":"c
 // ==== Challenge 1: Use .forEach() ====
 // The event director needs both the first and last names of each runner for their running bibs.  Combine both the first and last names into a new array called fullName. 
 let fullName = [];
+runners.forEach(element => fullName.push(`${element.first_name} ${element.last_name}`));
+//console.log(JSON.stringify(fullName))
 console.log(fullName);
 
 // ==== Challenge 2: Use .map() ====
 // The event director needs to have all the runner's first names converted to uppercase because the director BECAME DRUNK WITH POWER. Convert each first name into all caps and log the result
 let allCaps = [];
+allCaps = runners.map(element => element.first_name.toUpperCase());
 console.log(allCaps); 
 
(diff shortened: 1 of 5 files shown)


assignments/array-methods.js
```diff
@@ -56,28 +56,52 @@ const runners = [{"id":1,"first_name":"Charmain","last_name":"Seiler","email":"c
...
 console.log(largeShirts);
+console.log(largeShirtsNames);
 
 // ==== Challenge 4: Use .reduce() ====
 // The donations need to be tallied up and reported for tax purposes. Add up all the donations into a ticketPriceTotal array and log the result
+
+// Dedicated function
+{
+    let ticketPriceTotal = [];
+    ticketPriceTotal = runners.map(element => element.donation);
```
trevorfehrman, line 83: element is good as a semantic name but by definition we know it's an element.  wouldn't it be even better if we used something like "racer" or "runner" be our stand in for each element?

assignments/callbacks.js
```diff
@@ -20,33 +20,82 @@ const items = ['Pencil', 'Notebook', 'yo-yo', 'Gum'];
...
+//       return true;
+//     }
+//   })
+
+function comparerFor(list){
+  tmpArray = [];
+  for(i=0;i<list.length;i++){
+    if(tmpArray.indexOf(list[i]) > -1){
+      continue;
+    } else {
```
trevorfehrman, line 89: Yeah if you didn't figure it out yourself already I figured out why this works.  Index of returns -1 if the argument isn't in the array upon which it is invoked.  Includes(); is a most straightforward tool for this, or a Set.  This is totally valid though.

assignments/closure.js
```diff
@@ -1,14 +1,34 @@
 // ==== Challenge 1: Write your own closure ====
 // Write a simple closure of your own creation.  Keep it simple!
-
+hello = "Hello"
+function Greet(param){
```
trevorfehrman, line 4: Conventionally we reserve capital letters for Classes and Constructor functions in JS

trevorfehrman (commented):
Excellent job as usual.  Great work helping our classmates too.

PR #435, opened 2018-10-16
Jack Gipson

jackgipson:


Kamalnrf (commented):
Good work in array-methods, try finishing the remaining challenges in your free time.

assignments/callbacks.js
```diff
@@ -1,7 +1,9 @@
 // Create a callback function and invoke the function to test your work. You have been provided an example of a problem and a solution to see how this works with our items array.  Study both the problem and the solution to figure out the rest of the problems.
 
-const items = ['Pencil', 'Notebook', 'yo-yo', 'Gum'];
+const items = ['Pencil', 'Notebook', 'yo-yo', 'Gum', 'Lamp', 'Glass'];
 
+let lastIndex = items.length -1;
+console.log(items[lastIndex]);
```
ashwins93, line 6: Always remember that anywhere you have a number, you can replace it with a variable. You can even have an expression that will result in a number like `items[1 + 2 - 1 + 3 -2]`

assignments/callbacks.js
```diff
@@ -30,15 +32,31 @@ getLength(items, function(length) {console.log(length); });
 
 function last(arr, cb) {
   // last passes the last item of the array into the callback.
-}
+    let lastIndex = arr.length - 1;
+let lastWord = (arr[lastIndex]);
+cb(lastWord);
+  }
+last(items, function(crystal) {console.log(crystal); })
```
ashwins93, line 39: Hopefully this made callbacks "crystal" clear 😄

assignments/callbacks.js
```diff
@@ -30,15 +32,31 @@ getLength(items, function(length) {console.log(length); });
...
+
 
 function sumNums(x, y, cb) {
   // sumNums adds two numbers (x, y) and passes the result to the callback.
+  return cb(x, y);
```
ashwins93, line 44: We need to pass the result into the callback. 

```js
let sum = x + y;
return cb(sum);
```

assignments/callbacks.js
```diff
@@ -30,15 +32,31 @@ getLength(items, function(length) {console.log(length); });
...
 }
 
+const add = function(x, y) {
+  console.log(x + y);
+}
+sumNums(5, 2, add);
+
 function multiplyNums(x, y, cb) {
   // multiplyNums multiplies two numbers and passes the result to the callback.
+return cb(x, y);
```
ashwins93, line 54: Same as above. Pass the result onto the callback.

assignments/closure.js
```diff
@@ -1,14 +1,33 @@
...
+
+    function phlox() {
+      const sayHi = 'How do you do?';
+      console.log(`${hello} but I say ${sayHello} and I say ${sayHi}`)
+    }
+    phlox();
+  }
+  neelix();
+}
+spock('Live Long And Prosper');
```
ashwins93, line 19: Very good example here.

assignments/closure.js
```diff
@@ -1,14 +1,33 @@
...
 
 // ==== Challenge 2: Create a counter function ====
 const counter = () => {
   // Return a function that when invoked increments and returns a counter variable.
+  let count = 0;
+  return () => (++count);
```
ashwins93, line 25: That is very terse. Are you comfortable with arrow functions?

ashwins93 (commented):
You are definitely getting there. Refer to the solution code and remember that it gets better with repetitions. Keep it up.

PR #439, opened 2018-10-16
Daren Larson

darenlarson:


assignments/array-methods.js
```diff
@@ -56,28 +56,76 @@ const runners = [{"id":1,"first_name":"Charmain","last_name":"Seiler","email":"c
...
+    fullName.push(`${item.first_name} ${item.last_name}`);
+});
+
 console.log(fullName);
 
 // ==== Challenge 2: Use .map() ====
 // The event director needs to have all the runner's first names converted to uppercase because the director BECAME DRUNK WITH POWER. Convert each first name into all caps and log the result
 let allCaps = [];
+runners.map(function(item) {
+    allCaps.push(item.first_name.toUpperCase());
```
ashwins93, line 69: It is not quite how we would use a  `map` function. The callback to the map function should return a transformed element from the original element and the `map` function itself returns a new array filled with the transformed elements. This is how we would typically use it

```js
 let allCaps = runners.map(function (runner) {
  return runner.first_name.toUpperCase();
});
```

assignments/array-methods.js
```diff
@@ -56,28 +56,76 @@ const runners = [{"id":1,"first_name":"Charmain","last_name":"Seiler","email":"c
...
+const bigDonors = runners.filter((data) => {
+    return data.donation > 200;
+});
+
+console.log(bigDonors);
 
 // Problem 2
+// Add the first name, last name, and company name together.
+let NameCompany = [];
+runners.forEach(function(item) {
```
ashwins93, line 108: `map` may have better suited for this particular task.

assignments/callbacks.js
```diff
@@ -2,46 +2,80 @@
...
+
+
+// Question 5
 function contains(item, list, cb) {
   // contains checks if an item is present inside of the given array/list.
   // Pass true to the callback if it is, otherwise pass false.
-}
+  if(list.includes(item)) {
+    return cb(true);
+  } else cb(false);
```
ashwins93, line 73: The `else` is not necessary here as we are `return`ing inside `if` clause

assignments/closure.js
```diff
@@ -1,11 +1,32 @@
...
+parent();
+
 
 
 // ==== Challenge 2: Create a counter function ====
 const counter = () => {
   // Return a function that when invoked increments and returns a counter variable.
+  let count = 0;
+
+  return () => (++count);
```
ashwins93, line 20: Really good use of arrow function to write terse code.

ashwins93 (commented):
Really good job getting all the tasks done. Would really love to see you go for the stretch goals. Good job with the git commits.

PR #447, opened 2018-10-16
Kevin Marker

kmarker1101:


assignments/callbacks.js
```diff
@@ -24,25 +24,49 @@ const items = ['Pencil', 'Notebook', 'yo-yo', 'Gum'];
...
 }
 
+multiplyNums(2, 3, function (product) {
+    console.log(product)
+});
+
 function contains(item, list, cb) {
   // contains checks if an item is present inside of the given array/list.
   // Pass true to the callback if it is, otherwise pass false.
+    return cb(item in list);
```
johnoro: This doesn't work! Try to do it in a JavaScript type of way.

assignments/callbacks.js
```diff
@@ -24,25 +24,49 @@ const items = ['Pencil', 'Notebook', 'yo-yo', 'Gum'];
...
 }
 
+contains('cat', items, function (inList) {
+    console.log(inList)
+});
```
johnoro, line 69: You could forego the function here and use `console.log` as your callback:
```javascript
contains('cat', items, console.log);
```
since `log` is just a function (specifically, `log` is a method on the `console` object).

assignments/array-methods.js
```diff
@@ -56,28 +56,56 @@ const runners = [{"id":1,"first_name":"Charmain","last_name":"Seiler","email":"c
...
+});
+console.log(fullName);
 
-// Problem 3
\ No newline at end of file
+// Problem 3 What is the average donation per shirt size
+// let average = [];
+// runners.forEach(item => {
+//     average.push()
+// })
```
johnoro: Try to finish this up if you get the chance!

johnoro (commented):
Great job!

PR #448, opened 2018-10-16
Brandon Campbell

BrandCam:
Call backs hurt my brain, but ill get it by tomorrow.

### DIFF
--- a/assignments/callbacks.js
+++ b/assignments/callbacks.js
@@ -73,7 +73,7 @@ contains("Gum", items, function(bul){
   console.log(bul)
 }); */
 
-
+/*
 function cB(item, list) {
   let newArr = [];
   for (let i = 0; i < list.length; i++){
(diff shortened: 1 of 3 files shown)


assignments/callbacks.js
```diff
@@ -19,30 +19,87 @@ const items = ['Pencil', 'Notebook', 'yo-yo', 'Gum'];
...
 
 
 function getLength(arr, cb) {
-  // getLength passes the length of the array into the callback.
+   return cb(arr.length);
 }
-
+getLength(items, function(length){
+  console.log(length)
+});
```
johnoro, line 30: You could forego this extra function wrapping `console.log` by just using `log` as your callback function (and this applies elsewhere in your code, too):
```javascript
```
This works since `log` is just a function; specifically it is a method on the `console` object.

johnoro (commented):
Awesome work; try to attempt some of the stretch!

PR #449, opened 2018-10-16
David Freitag

freitagdavid:


assignments/callbacks.js
```diff
@@ -21,32 +24,58 @@ const items = ['Pencil', 'Notebook', 'yo-yo', 'Gum'];
...
-
 function getLength(arr, cb) {
-  // getLength passes the length of the array into the callback.
+    // getLength passes the length of the array into the callback.
+    return cb(arr.length);
 }
 
+getLength(items, function(length) {
+    console.log(length);
+});
```
johnoro, line 34: You could forego the wrapping function and just use `console.log` as your callback here:
```javascript
```
then, because of `return cb(arr.length)` will pretty much equate to `console.log(items.length)`

assignments/array-methods.js
```diff
@@ -2,82 +2,538 @@
...
-// Problem 3
\ No newline at end of file
+let donors = runners.map(item => item.company_name);
+donors.forEach(
+    company =>
+        (donors[company] = runners
+            .filter(item => item.company_name === company)
+            .map(item => item.donation)
+            .reduce((total, item) => total + item))
+);
```
johnoro, line 538: Here is an example of another approach to this problem:
```javascript
const donationsByCompany = runners
  .reduce((acc, { company_name, donation }) => {
    if (acc.hasOwnProperty(company_name)) {
      acc[company_name] += donation;
      return acc;
    } else {
      return { ...acc, [company_name]: donation };
    }
  }, {});
```
using [object destructuring](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment#Object_destructuring) and [spread syntax](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax#Spread_in_object_literals).

johnoro (commented):
Great work!

PR #452, opened 2018-10-16
Zayna Crick

ZaynaCrick:


assignments/array-methods.js
```diff
@@ -56,28 +56,60 @@ const runners = [{"id":1,"first_name":"Charmain","last_name":"Seiler","email":"c
...
+    return large.shirt_size === "L";
+})
 console.log(largeShirts);
 
 // ==== Challenge 4: Use .reduce() ====
 // The donations need to be tallied up and reported for tax purposes. Add up all the donations into a ticketPriceTotal array and log the result
-let ticketPriceTotal = [];
+
+const ticketPriceTotal = runners.reduce(function(donate,money) {
+    return donate + money.donation;
```
trevorfehrman, line 87: try more semantic names for parameters.  The first parameter will represented an accumulating value, so you could try something like "total".  the seoncd is each object in the runners array so you might try "runner" or "racer."

To be clear you can name these w/e you want, but that means you have the opportunity to make them very descriptive, helps with remember what they do.

assignments/closure.js
```diff
@@ -1,14 +1,40 @@
...
 // Write a simple closure of your own creation.  Keep it simple!
+// Closures 101
+
+function PetsName(name) {
+  const Pet = name;
+  const Behavior ="is well behaved."
+  console.log(`${Pet} ${Behavior}`);
+  debugger;
+
+  function PetInfo() {
```
trevorfehrman, line 11: This looks good, good grasp of the subject.

trevorfehrman (commented):
GJ Zayna!  Putting this all together

PR #454, opened 2018-10-16
Max David Metelus: "mvp- complete"

Mdmetelus:
minimum viable product.

### DIFF
--- a/assignments/array-methods.js
+++ b/assignments/array-methods.js
@@ -56,11 +56,11 @@ const runners = [{"id":1,"first_name":"Charmain","last_name":"Seiler","email":"c
 // ==== Challenge 1: Use .forEach() ====
 // The event director needs both the first and last names of each runner for their running bibs.  Combine both the first and last names into a new array called fullName.
 let fullName = [];
-// runners.foreach(` ${ runners[i]["first_name"] } ${ runners[i]["last_name"] }`);
+runners.foreach(` ${ runners[i]["first_name"] } ${ runners[i]["last_name"] }`);
 
-// runners.foreach(fullName.push(runners[i].first_name + " " + runners[i].last_name ));
-//
-// console.log(fullName);
+runners.foreach(fullName.push(runners[i].first_name + " " + runners[i].last_name ));
+
+console.log(fullName);
 
 // ==== Challenge 2: Use .map() ====
 // The event director needs to have all the runner's first names converted to uppercase because the director BECAME DRUNK WITH POWER. Convert each first name into all caps and log the result
--- a/assignments/array-methods.js
+++ b/assignments/array-methods.js
@@ -54,30 +54,46 @@ const runners = [{"id":1,"first_name":"Charmain","last_name":"Seiler","email":"c
 {"id":50,"first_name":"Shell","last_name":"Baine","email":"sbaine1d@intel.com","shirt_size":"M","company_name":"Gabtype","donation":171}];
 
 // ==== Challenge 1: Use .forEach() ====
-// The event director needs both the first and last names of each runner for their running bibs.  Combine both the first and last names into a new array called fullName. 
+// The event director needs both the first and last names of each runner for their running bibs.  Combine both the first and last names into a new array called fullName.
 let fullName = [];
-console.log(fullName);
+// runners.foreach(` ${ runners[i]["first_name"] } ${ runners[i]["last_name"] }`);
+
+// runners.foreach(fullName.push(runners[i].first_name + " " + runners[i].last_name ));
+//
+// console.log(fullName);
 
 // ==== Challenge 2: Use .map() ====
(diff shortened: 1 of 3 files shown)


assignments/array-methods.js
```diff
@@ -54,30 +54,46 @@ const runners = [{"id":1,"first_name":"Charmain","last_name":"Seiler","email":"c
 {"id":50,"first_name":"Shell","last_name":"Baine","email":"sbaine1d@intel.com","shirt_size":"M","company_name":"Gabtype","donation":171}];
 
 // ==== Challenge 1: Use .forEach() ====
-// The event director needs both the first and last names of each runner for their running bibs.  Combine both the first and last names into a new array called fullName. 
+// The event director needs both the first and last names of each runner for their running bibs.  Combine both the first and last names into a new array called fullName.
 let fullName = [];
+runners.foreach(` ${ runners[i]["first_name"] } ${ runners[i]["last_name"] }`);
```
szincone, line 59: You're mixing up forEach `arr.forEach(element => element` and for loops `for (let i=0; i < arr.length; i++)`. They do the same thing but you can only use one or the other.

assignments/array-methods.js
```diff
@@ -54,30 +54,46 @@ const runners = [{"id":1,"first_name":"Charmain","last_name":"Seiler","email":"c
...
+
+runners.foreach(fullName.push(runners[i].first_name + " " + runners[i].last_name ));
```
szincone, line 61: The inside you made `fullName.push(runners[i].first_name + " " + runners[i].last_name` is correct if you were trying to do for loops(we weren't trying to do for loops for this assignment though), you just would have to had put it in a for loop like this `for (let i = 0; i < runners.length;i++){
  fullName.push(runners[i].first_name + " " + runners[i].last_name )
}`

assignments/array-methods.js
szincone: This line should have looked something like ```runners.forEach( runner => {
    fullName.push(`${runner.first_name} ${runner.last_name}`);
})``` the forEach just goes through the runners array for us. Then inside you do your action. So now full-name would have our first and last names.

assignments/array-methods.js
```diff
@@ -54,30 +54,46 @@ const runners = [{"id":1,"first_name":"Charmain","last_name":"Seiler","email":"c
...
 // The event director needs to have all the runner's first names converted to uppercase because the director BECAME DRUNK WITH POWER. Convert each first name into all caps and log the result
 let allCaps = [];
-console.log(allCaps); 
+allCaps = runners.map(function(items) {return items.first_name.toUpperCase();});
+console.log(allCaps);
 
 // ==== Challenge 3: Use .filter() ====
 // The large shirts won't be available for the event due to an ordering issue.  Get a list of runners with large sized shirts so they can choose a different size. Return an array named largeShirts that contains information about the runners that have a shirt size of L and log the result
 let largeShirts = [];
+largeShirts = runners.filter((list) => {return list.shirt_size === "L" ;});
```
szincone, line 74: `runners.filter((list) => {return list.shirt_size === "L" ;});` could also be written as `runners.filter(list => list.shirt_size === "L" );` (you don't need the return statement if it's written on the same line because it's implied. Parentheses are also optional on an arrow function if you're only passing 1 param.

assignments/array-methods.js
```diff
@@ -54,30 +54,46 @@ const runners = [{"id":1,"first_name":"Charmain","last_name":"Seiler","email":"c
...
+ticketPriceTotal = runners.reduce((total, next) => {return total + next.donation ;});
 console.log(ticketPriceTotal);
 
 // ==== Challenge 5: Be Creative ====
 // Now that you have used .forEach(), .map(), .filter(), and .reduce().  I want you to think of potential problems you could solve given the data set and the 5k fun run theme.  Try to create and then solve 3 unique problems using one or many of the array methods listed above.
 
-// Problem 1
+// Problem 1 - MAKE A NICKNAME FOR EACH CONTESTEANT USING THIER FIRST NAM AND SHIRT SIZE.
+let nickname =[];
+nickname = runners.map(items.first_name.toUpperCase.concat(items.shirt_size));
```
szincone, line 88: - Checkout the syntax on MDN https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/map. 
- Map requires a callback. Should be something like `runners.map(items => items.first_name.toUpperCase.concat(items.shirt_size))`. (This won't run either because your missing you parentheses on toUpperCase, it should be `toUpperCase()`. It will work after that though)
- I would also change `items` to `item` or `runner` since it represents a single runner. (This is a really cool way of using concat to do it, first person I've seen using it)

szincone (commented):
- Your thought process was right on most of these, just syntax errors for the most part. Interesting and unique ways to solve some of your challenges (like the nickname one)

PR #459, opened 2018-10-17
final

alexcartaz:


assignments/callbacks.js
```diff
@@ -19,30 +20,64 @@ const items = ['Pencil', 'Notebook', 'yo-yo', 'Gum'];
...
+//----------------------------------
 
 function getLength(arr, cb) {
   // getLength passes the length of the array into the callback.
+  return cb(arr.length);
 }
 
+getLength(items, function(length) {
+  console.log(length);
+});
```
johnoro, line 33: You could forego the extra function and just use the `log` method on the `console` object as your callback function:
```javascript
```

assignments/callbacks.js
```diff
@@ -19,30 +20,64 @@ const items = ['Pencil', 'Notebook', 'yo-yo', 'Gum'];
...
+multiplyNums(1, 2, function(product) {
+  console.log(product);
+});
+
+//----------------------------------
+
 function contains(item, list, cb) {
   // contains checks if an item is present inside of the given array/list.
   // Pass true to the callback if it is, otherwise pass false.
+  return cb( ( list.indexOf(item) < 0 ) ? false : true );
```
johnoro, line 74: You could shorten this to:
```javascript
return cb( list.indexOf(item) > -1 );
```
or
```javascript
return cb( list.indexOf(item) !== -1 );
```
or using the array method [`includes`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/includes) for brevity and clarity:
```javascript
return cb( list.includes(item) );
```
since all of these will evaluate to `true` or `false`.

assignments/callbacks.js
johnoro: This goes for your other callbacks using this unneeded pattern too!

assignments/closure.js
```diff
@@ -1,14 +1,34 @@
...
+//function counter(){
 const counter = () => {
-  // Return a function that when invoked increments and returns a counter variable.
-};
-// Example usage: const newCounter = counter();
-// newCounter(); // 1
-// newCounter(); // 2
+	var c = 1;
+	function increment(){
+		return c++;
```
johnoro, line 23: It's interesting to use a post-increment here; it works, but it almost feels misleading. I assume it was intentional, but just keep in mind that you could've done it with a with a pre-increment:
```javascript
const counter = () => {
  let c = 0;
  const increment = () => ++c;
  return increment;
};
```

johnoro (commented):
Great work! You forgot the fifth challenge in `array-methods.js`, though.

PR #485, opened 2018-10-20
Phillip Smith

smithpb:
Submission of the project for the Javascript II project.  Completed all requirements and the stretch goals presented.

@keirankozlowski

### DIFF
--- a/assignments/array-methods.js
+++ b/assignments/array-methods.js
@@ -56,28 +56,46 @@ const runners = [{"id":1,"first_name":"Charmain","last_name":"Seiler","email":"c
 // ==== Challenge 1: Use .forEach() ====
 // The event director needs both the first and last names of each runner for their running bibs.  Combine both the first and last names into a new array called fullName. 
 let fullName = [];
+runners.forEach(function(element) {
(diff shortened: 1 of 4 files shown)


keirankozlowski:
### Objective

### MVP
**Objectives**

- [x] Array Methods MVP
- [x] Callbacks MVP
- [x] Closure MVP

### Stretch
**Goals**

- [x] Callbacks stretch
- [x] Closure stretch
- [x] ES6 arrow syntax stretch

### Great
- Nice use of template literals for pushing items into arrays and consoling.
- Very creative with your Array Methods Challenge 5 problems.
- Your Callbacks section looks great!

### Requests for Improvement
- For the Array Methods Challenge 4, no need to return as an object. Returning a number would have been fine.
- Small bug in your Callbacks section in the "contains" function. Check out this article: https://codeburst.io/javascript-double-equals-vs-triple-equals-61d4ce5a121a
- To be a closure, it has to have an enclosing function.

### Commits
- Commit more frequently. At the very least, you should have committed at least per each file completion.

### Rating: 3

PR #487, opened 2018-10-20
Javascript II

acarmstrong98:


abravebee:
# Objectives
- With some basic JavaScript principles in hand, we can now expand our skills out even further by exploring callback functions, array methods, and closure. Finish each task in order as the concepts build on one another. 

# MVP
## Task 1: Set Up The Project With Git

- [x]  Fork the project into your GitHub user account
- [x]  Clone the forked project into a directory on your machine
- [x]  You are now ready to build this project with your preferred IDE
- [x]  To test your console.log() statements, open up the index.html file found in the assignments folder and use the developer tools to view the console.

## Task 2: Callbacks
This task focuses on getting practice with callback functions by giving you an array of values and instructions on what to do with that array.

- [x] Review the contents of the callbacks.js file. Notice you are given an array at the top of the page. Use that array to aid you with your callback functions.
- [x] Write out each function using the ES5 function keyword syntax.
- [ ] Solve the problems listed. Save the stretch problems until you have completed Tasks 1-4.

## Task 3: Array Methods

- [x] Use .forEach(), .map(), .filter(), and .reduce() to loop over an array with 50 objects in it. The array-methods.js file contains several challenges built around a fundraising 5K fun run event.
- [x]  Review the contents of the array-methods.js file.
- [ ]  Complete the problems provided to you
- [ ] Notice the last three problems are up to you to create and solve. This is an awesome opportunity for you to push your critical thinking about array methods, have fun with it.

## Task 4: Closures
- [ ] We have learned that closures allow us to access values in scope that have already been invoked.
- [ ] Hint: Utilize debugger statements in your code in combination with your developer tools to easily identify closure values.
- [ ]  Review the contents of the closure.js file.
- [ ]  Solve the problems listed. Save the stretch problems until you have completed Tasks 1-4.
- [ ]  Once you have completed this task please submit a pull request against the original fork.

# Stretch Goals
- [ ] Arrow Function Syntax - Check out this awesome guide for ES6 arrow syntax. You will see more and more arrow functions as you progress deeper into JavaScript. Use the stretch-function-conversion.js file as a helper challenge to showcase some of the differences between ES5 and ES6 syntax.
- [ ]  Look up what an IIFE is in JavaScript and experiment with them

# Great
- You're very very close to getting the callbacks and array methods right, but need a little more practice, I think!

# Requests for Improvement
- `function` is mistyped as `funtion` several times, which is throwing errors.
- Look again at the callbacks example. We need to pass something into each `cb` in the outer function, and make sure we're returning or console.logging something inside the callback function. The name of each outer function gives you a clue of what should be passed into `cb`. There's a pattern to be extrapolated from the given example--see if you can replicate it in your answers. You've got a lot of the right pieces; they just need to be rearranged.
- On Array Methods Challenge 1, the `fullName` array is what we want to push our first and last names on to; it doesn't need to be passed anywhere. You've got the right idea with `item.first_name` and `item.last_name`, but if you check your console, the program doesn't know what `item` is. This is actually what you want as your param instead of `fullName`. 
- On Challenge 2, you're trying to apply `toUppercase` to every object in `runners` instead of just the first names from each object. 
- On Challenge 3, there is no property called `L` but some objects have the value of `L` set to one of their properties.
- Make sure you're checking all of your work in the console! I highly recommend the Quokka extension if you are using VSCode, as it will give you results for console.logs in-line in the text editor, right next to each console.log! It makes it much easier to see and catch the types of bugs you're running into.

# Rating: 1

PR #490, opened 2018-10-20
Evan Leduc

evan-leduc:
@keirankozlowski
Finished MVP.
Questions:
1. Nvm had a breakthrough regarding challenge 2 in closure.js. I was thinking that the x variable was deleted after the function was called, just like in a loop. So the variable exists regardless if the function is called, and it is referenced through closure when I call the new function.
2. In challenge 4 in array-methods.js is there a way in which you can store the reduce value straight to the new array as opposed to the way I did it.

### DIFF
--- a/assignments/array-methods.js
+++ b/assignments/array-methods.js
@@ -89,7 +89,25 @@ console.log(ticketPriceTotal);
 // Now that you have used .forEach(), .map(), .filter(), and .reduce().  I want you to think of potential problems you could solve given the data set and the 5k fun run theme.  Try to create and then solve 3 unique problems using one or many of the array methods listed above.
 
 // Problem 1
+// Get a list of needed shirt sizes
+let shirtList = runners.map(function(runner){
+    return runner.shirt_size;
+})
+console.log(shirtList.sort());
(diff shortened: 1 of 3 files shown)


keirankozlowski:
### Objective

### MVP
**Objectives**

- [x] Array Methods MVP
- [ ] Callbacks MVP
- [x] Closure MVP

### Stretch
**Goals**

- [ ] Callbacks stretch
- [ ] Closure stretch
- [ ] ES6 arrow syntax stretch

### Great
- Nice job being creative with the Array Methods Challenge 5 problems.
- Counter looks great!
- Your code is always nice to read. Keep it organized like this all the time!

### Requests for Improvement
- For the Callbacks assignment, look at the example problem. Each of your solutions is missing the second part! 
- For the Closures assignment challenge 1, you should call your internal function after its declaration.

### Commits
- Keep working on committing more frequently. :)

### Rating: 1
If you fix the Callbacks section issue, I will upgrade this to a 2. 

Answer to your second question: check out the model response that I linked in Slack tonight. The example does everything in a single function. :)

PR #499, opened 2018-10-20
Completed MVP and just about all stretch goals. 

jrizza88:
Had trouble converting forEach functions to arrow function styles. Other than that, mostly everything is ES6


jrizza88:
@Austinlp4

PR #501, opened 2018-10-20
Andrew Ogle

andrewogle:
I committed more than whats here but I tried to do a multi line comment and I think I did it wrong because I do not see me earlier commit messages.
@keirankozlowski

### DIFF
--- a/assignments/array-methods.js
+++ b/assignments/array-methods.js
@@ -56,11 +56,17 @@ const runners = [{"id":1,"first_name":"Charmain","last_name":"Seiler","email":"c
 // ==== Challenge 1: Use .forEach() ====
 // The event director needs both the first and last names of each runner for their running bibs.  Combine both the first and last names into a new array called fullName. 
 let fullName = [];
+runners.forEach(element =>{
+    fullName.push(`${element.first_name} ${element.last_name}`)
+})
 console.log(fullName);
 
 // ==== Challenge 2: Use .map() ====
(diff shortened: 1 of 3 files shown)


keirankozlowski:
### Objective

### MVP
**Objectives**

- [ ] Array Methods MVP
- [x] Callbacks MVP
- [x] Closure MVP

### Stretch
**Goals**

- [x] Callbacks stretch
- [ ] Closure stretch
- [ ] ES6 arrow syntax stretch

### Great
- Array Methods Challenge 1, nice use of template literals!
- Array Methods nice job using the arrow syntax and dot chaining your functions!
- Nice job using anon functions in your Callbacks assignment.
- Good job with closures! This section looks solid.

### Requests for Improvement
- Don't forget to use the Format Document command prior to submission to clean up your code.
- Please finish Array Methods Challenge 5!
- For the Callbacks assignment with the "contains" function, using the forEach loop actually misses the mark here. You will return a value for each item, instead of returning true or false once depending on if the array contains the given element. Re-implement this as a simple for loop.
- For your Closures assignment with the counter function, use return instead of console.log to give the newCount value. You could also use an increment operator (x++) instead of (x + 1), but your implementation works.

### Commits
- You're improving at commits, keep it up!

### Rating: 1
I had to give you a 1 because the Array Methods Challenge 5 was incomplete; once you finish that, it'll be a 3.

PR #522, opened 2018-11-27
Kelli Hanson - JavaScript-II

kelliraehanson:
@VivaCode

### DIFF
--- a/assignments/array-methods.js
+++ b/assignments/array-methods.js
@@ -90,24 +90,29 @@ largeShirts = runners.filter(function(shirts) {
 console.log(largeShirts);
 
 // ==== Challenge 4: Use .reduce() ====
-// The donations need to be tallied up and reported for tax purposes. Add up all the donations into a ticketPriceTotal array and log the result
+// The donations need to be tallied up and reported for tax purposes. 
+//Add up all the donations into a ticketPriceTotal array and log the result
 // let ticketPriceTotal = [];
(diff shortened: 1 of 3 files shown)


fivegoblins:
**Project Objectives**

- **Callbacks**: You don't seem to be using callbacks correctly in this challenge. I'd advise that you go back into the training kit and study callback functions.

- **Array methods**: You did not use the instructed array methods in this challenge. You will need to get practice using .map(), .filter(), and .reduce(), as well as .forEach().

- **Closures**: The closure that you wrote yourself in this challenge works successfully, but you did not complete the counter exercise.

I would recommend re-watching the lecture video from this day and then attempting to re-factor this project so that you can be prepared for Friday's sprint challenge.

**Rating: 1**

VivaCode (commented):
#MVP

- [x] Callbacks
- [x] Array Methods (Use .forEach(), .map(), .filter(), and .reduce())
- [x] Closures

_Kelli, good job on refactoring this. I think your understanding of callbacks is getting stronger. On ln 46 on callbacks. use .length-1 to get the last item no matter how many items are on the list. Your solution for function contains was very interesting. I learned something new (trueFalse). Good job! Keep pushing through!_

#STRETCH

- [x] Arrow Function Syntax ES5 to ES6
- [ ] Write some practice IFEEs
_I saw a few functions written as arrows, I think you have a good feel for them_

#COMMITS
good!

#RATING: 3

PR #580, opened 2018-11-28
Jason-Holloway callbacks

jjamaltwin:
@vivacode

### DIFF
--- a/assignments/callbacks.js
+++ b/assignments/callbacks.js
@@ -21,22 +21,42 @@ const items = ['Pencil', 'Notebook', 'yo-yo', 'Gum'];
 
 */
 
-
-function getLength(arr, cb) {
-  // getLength passes the length of the array into the callback.
+// getLength passes the length of the array into the callback.
+ 
+function getLength(arr, cb) {  return cb(arr.length);
 }
+getLength(items, function(length) {
+ console.log(length);
+});
+
 
 function last(arr, cb) {
-  // last passes the last item of the array into the callback.
+  return cb(arr[arr.length-1]);
 }
+  last(items, log);
+
+  // last passes the last item of the array into the callback.
+
 
 function sumNums(x, y, cb) {
-  // sumNums adds two numbers (x, y) and passes the result to the callback.
+// sumNums adds two numbers (x, y) and passes the result to the callback.
+function sumNums(x, y, cb){
+  return cb(x+y);
 }
+sumNums(2, 4, log)
+
+}
+
+return cb(x+y);
 
 function multiplyNums(x, y, cb) {
   // multiplyNums multiplies two numbers and passes the result to the callback.
 }
+function multiplyNums(x, y, cb) {
+  return cb(x*y);
+ }
+   multiplyNums(4,4, log)
+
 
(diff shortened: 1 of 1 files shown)


assignments/callbacks.js
```diff
@@ -21,22 +21,42 @@ const items = ['Pencil', 'Notebook', 'yo-yo', 'Gum'];
...
+getLength(items, function(length) {
+ console.log(length);
+});
+
 
 function last(arr, cb) {
-  // last passes the last item of the array into the callback.
+  return cb(arr[arr.length-1]);
 }
+  last(items, log);
```
VivaCode, line 36: Unfortunately, this doesn't return for you bc you don't have the same log function that Matt used. 
You can make this work by doing exactly what you did above and making it call items and a function.

last(items, function(lastItem) {
//   console.log(lastItem);
// });

assignments/callbacks.js
```diff
@@ -21,22 +21,42 @@ const items = ['Pencil', 'Notebook', 'yo-yo', 'Gum'];
 
 */
 
-
-function getLength(arr, cb) {
-  // getLength passes the length of the array into the callback.
+// getLength passes the length of the array into the callback.
+ 
+function getLength(arr, cb) {  return cb(arr.length);
```
VivaCode, line 26: This is good, just make sure to put return on the next line
solution:
function getLength(arr, cb) {
//   // getLength passes the length of the array into the callback.
//   console.log(cb(arr));
// }

// let arrayLength = function(items) {
//   return items.length;
// }

// getLength(items,arrayLength);

assignments/callbacks.js
```diff
@@ -21,22 +21,42 @@ const items = ['Pencil', 'Notebook', 'yo-yo', 'Gum'];
...
+
+return cb(x+y);
 
 function multiplyNums(x, y, cb) {
   // multiplyNums multiplies two numbers and passes the result to the callback.
 }
+function multiplyNums(x, y, cb) {
+  return cb(x*y);
+ }
+   multiplyNums(4,4, log)
```
VivaCode, line 58: these are all the same, using log I think twisted you up. 
https://codeburst.io/javascript-what-the-heck-is-a-callback-aba4da2deced
https://www.geeksforgeeks.org/javascript-callbacks/

VivaCode (commented):
# MVP

#### set up project
- [x] Fork the project into your GitHub user account
- [x] Clone the forked project into a directory on your machine
- 
# Objectives

_Callbacks are tough. They take lots of practice and reading several docs before it clicks. Don't give up. I know you can do this! I added a couple docs in the notes that might be helpful. It's not easy to ask for help, but we all want to see you succeed. If I don't explain it in a way that's helpful, I'll find someone else. Don't be an island._

- [ ] Callbacks
- [ ] Write out each function using the ES5 function keyword syntax.
- [ ] Array Methods Use .forEach(), .map(), .filter(), and .reduce() to loop over an array with 50 objects in it.
- [ ] Create and solve the last three problems
- [ ] Closures

# Stretch

- [ ] Arrow Function Syntax 
- [ ] Stretch  IIFE

# Commits


# Rating: 1

PR #616, opened 2019-01-29
Talent async

sampler36:
New and better callbacks and array methods with closures

### DIFF
--- a/assignments/callbacks.js
+++ b/assignments/callbacks.js
@@ -1,7 +1,5 @@
 // Create a callback function and invoke the function to test your work. You have been provided an example of a problem and a solution to see how this works with our items array.  Study both the problem and the solution to figure out the rest of the problems.
 
-const items = ['Pencil', 'Notebook', 'yo-yo', 'Gum'];
-
 /* 
 
   //Given this problem: 
@@ -20,18 +18,23 @@ const items = ['Pencil', 'Notebook', 'yo-yo', 'Gum'];
   });
 
 */
+const items = [1, 2, 3, 4, 'hello'] //with 5 elements where count starts from 0 || 
+        function firstItem(arr, cb) {
+          return cb(arr[4]);  // consoles hello
+        }
+        firstItem(items, function(first) {
+          console.log(first)
+        });
 
+const numero = [1, 2, 3, 4, 'hello'] //with 5 elements where count starts from 0 || 
 function getLength(arr, cb) {
-  
   return cb(arr.length)
-
   // getLength passes the length of the array into the callback.
 }
+console.log ();
 
 function last(arr, cb) {
-    // items[0] ; how to get the first item
      return cb(arr[arr.length -1])
-
   // last passes the last item of the array into the callback.
 }
 
@@ -47,6 +50,7 @@ function sumNums(x, y, cb) {
 
   return cb(arr[x + y]);
 }
+console.log(sumNums(5, 7));
 
 function multiplyNums(x, y, cb) {
   // multiplyNums multiplies two numbers and passes the result to the callback.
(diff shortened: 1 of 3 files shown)


assignments/closure.js
```diff
@@ -1,20 +1,48 @@
...
-const counter = () => {
   // Return a function that when invoked increments and returns a counter variable.
-};
 // Example usage: const newCounter = counter();
-// newCounter(); // 1
-// newCounter(); // 2
+const counter = () => {
+ let  counter  = 0;
+          return function newCounter(){
+            ++counter;
```
Velsu, line 18: You forgot to add `return` inside the function. Without it won't increment properly because the function is not returning anything.

assignments/closure.js
```diff
@@ -1,20 +1,48 @@
...
 /* STRETCH PROBLEM, Do not attempt until you have completed all previous tasks for today's project files */
 
 // ==== Challenge 3: Create a counter function with an object that can increment and decrement ====
-const counterFactory = () => {
   // Return an object that has two methods called `increment` and `decrement`.
   // `increment` should increment a counter variable in closure scope and return it.
+  // const arr = newParamArr.map((value, index) => ({[value]: paramArr[index]}))
+  // Increment
+
+      let a = 1;
```
Velsu, line 35: These ones are wrong as they are not even a functions.

assignments/callbacks.js
```diff
@@ -20,33 +18,70 @@ const items = ['Pencil', 'Notebook', 'yo-yo', 'Gum'];
...
+const items = [1, 2, 3, 4, 'hello'] //with 5 elements where count starts from 0 || 
+        function firstItem(arr, cb) {
+          return cb(arr[4]);  // consoles hello
+        }
+        firstItem(items, function(first) {
+          console.log(first)
+        });
 
-
+const numero = [1, 2, 3, 4, 'hello'] //with 5 elements where count starts from 0 ||
```
Velsu, line 29: ?????

assignments/callbacks.js
```diff
@@ -20,33 +18,70 @@ const items = ['Pencil', 'Notebook', 'yo-yo', 'Gum'];
...
   // sumNums adds two numbers (x, y) and passes the result to the callback.
+//   let a = 3
+// function addTwo(x) {
+//   let ret = x + 2
+//   return ret
+//  }
+//  let b = addTwo(a)
+//  console.log(b);
+
+  return cb(arr[x + y]);
```
Velsu, line 44: You want to return a callback that will add two numbers here, what You are doing is getting index of `arr` which is not even passed as an argument or defined inside the function.

assignments/callbacks.js
```diff
@@ -20,33 +18,70 @@ const items = ['Pencil', 'Notebook', 'yo-yo', 'Gum'];
...
 }
+console.log(sumNums(5, 7));
 
 function multiplyNums(x, y, cb) {
   // multiplyNums multiplies two numbers and passes the result to the callback.
+  return cb(arr[x * y]);
```
Velsu, line 50: Same as with sum, but You want to multiply here.

assignments/callbacks.js
```diff
@@ -20,33 +18,70 @@ const items = ['Pencil', 'Notebook', 'yo-yo', 'Gum'];
...
 }
 
 function contains(item, list, cb) {
   // contains checks if an item is present inside of the given array/list.
   // Pass true to the callback if it is, otherwise pass false.
+  const  contains = list.length;
```
Velsu: You are not calling passed `cb` inside the function.

Velsu (commented):
Antonio You got array-methods and closures properly, just with some small bugs like forgetting to return a value.
The biggest issues are in Your callback exercises. This is definetely one thing You need to focus on and get better understanding. It might look a little intimidating at first, but this is not really so hard as it looks. You are calling a functions, passing arguments and one of them is another function (cb). Then You just return this function (cb) as a callback passing in supplied arguments inside it. What to do with arguments You decide based on the functions, if it is `sum` then You add etc.

Also You forgot about our new git workflow in Lambda and You should be making this pull request to Your own repo.

Overall good job i can see You are improving.

PR #620, opened 2019-02-01
Myasia barbee Javascript-II

MMBarbee:


assignments/array-methods.js
```diff
@@ -58,11 +58,20 @@ const runners = [{"id":1,"first_name":"Charmain","last_name":"Seiler","email":"c
 let fullName = [];
 console.log(fullName);
 
+runners.forEach((getNames)=>{
```
lolax, line 61: Nice work!

assignments/array-methods.js
```diff
@@ -58,11 +58,20 @@ const runners = [{"id":1,"first_name":"Charmain","last_name":"Seiler","email":"c
...
+    fullName.push(names)
+  });
+
 // ==== Challenge 2: Use .map() ====
 // The event director needs to have all the runner's first names converted to uppercase because the director BECAME DRUNK WITH POWER. Convert each first name into all caps and log the result
 let allCaps = [];
 console.log(allCaps); 
 
+const caps = runners.map((capitalize));{
+    caps.push(capitalize.first_name.toUpperCase())
```
lolax: You don't need to push to the array caps because it's the array that is being returned by the map. You can instead write:
```suggestion
   const caps = runners.map((capitalize));{
     return capitalize.first_name.toUpperCase()
   }
```
Whatever you put as the return statement for the map will be added to the new array that map creates, which is caps in this case.

assignments/closure.js
```diff
@@ -1,15 +1,31 @@
 // ==== Challenge 1: Write your own closure ====
 // Write a simple closure of your own creation.  Keep it simple!
-
+const nickname = 'Trinkie';
```
lolax: Nice work with this closure!

lolax (commented):
Myasia, uou did a nice job on the array methods you worked on and the closures, but I see that you didn't quite finish this assignment. If you're struggling with a concept or problems in particular, please reach out so that I can help you work through it.

